PR #19, opened 2021-02-18
OpenShift Chart Compatibility

isutton:


Avni-Sharma:
Hi @isutton . How is the OpenShift Chart Compatibility different from Chart Compatibility on any other vanilla k8s?

mmulholla (approved):

Issue #48, opened 2021-03-11
Update chart verifier to use verify instead of certify


jasperchui:
Use verify instead of certify to align with tooling name

> chart-verifier certify --disable is-helm-v3 https://www.example.com/chart.tgz

Readme update required.

PR #49, opened 2021-03-16
Change command option verify to certify

mmulholla:
Update:
chart-verifier verify

To:
chart-verifier certify

see: https://github.com/redhat-certification/chart-verifier/issues/48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /.idea/
+/out/
+docker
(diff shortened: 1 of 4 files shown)


baijum:
Looks like there is confusion about the sub-command name. The title of #48 says to use `verify` but the example given there uses `certify`. Sometimes back I changed `certify` to `verify` in PR #17 based on issue #14.

mmulholla:
> Looks like there is confusion about the sub-command name. The title of #48 says to use `verify` but the example given there uses `certify`. Sometimes back I changed `certify` to `verify` in PR #17 based on issue #14.

oops, looks like i misread the issue.

PR #54, opened 2021-03-22
add makefile and improve github actions

mmulholla:
Add makefile with standard build tasks

Add go fmt and go mod tidy to build

Extend github actions to be used on Pull requests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,23 @@
+
+default: bin
+
+.PHONY: all
+all:  gomod_tidy gofmt bin test
+
+.PHONY: gomod_tidy
+gomod_tidy:
+	 go mod tidy
+
+.PHONY: gofmt
+gofmt:
+	go fmt -x ./...
+
+.PHONY: bin
+bin:
+	 hack/build.sh
+
+.PHONY: test
+test:
+	 go test -v ./...
+
+
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,6 +1,10 @@
 name: Build artifacts
 
-on: push
+on:
+    push:
+        branches: [ main ]
+    pull_request:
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,24 @@
...
+	go fmt -x ./...
+
+.PHONY: bin
+bin:
+	 go build *.go
+	 hack/build.sh
+
+.PHONY: test
+test:
+	 go test ./...
```
baijum, line 22: Can we add `-v` flag for verbose output?

mmulholla: Added.

mmulholla:
Note the build will not pass until this PR is merged: https://github.com/redhat-certification/chart-verifier/pull/51

Makefile
```diff
@@ -0,0 +1,24 @@
...
+	 go mod tidy
+
+.PHONY: gofmt
+gofmt:
+	go fmt -x ./...
+
+.PHONY: bin
+bin:
+	 go build *.go
+	 hack/build.sh
```
baijum, line 18: Do we need both these commands? Now there are two binaries, right?

mmulholla:
@baijum all working now - ready for final review

Makefile
mmulholla: True, removed the first.

.gitignore
```diff
@@ -1,2 +1,4 @@
 /.idea/
 /out/
+main
```
baijum, line 3: This may not be required as the binary output goes inside the `out` directory.

baijum:
@mmulholla The title still has "DO NOT MERGE". Please change it it's changed.

.gitignore
mmulholla: removed

baijum (approved):
LGTM

PR #58, opened 2021-03-25
add reports including chart.yaml content

mmulholla:
First pass at generating a report (everything could still change) - next steps are to add other info as requested, add some tests, and update the readme with instructions. 

Note the report is not generate if the tool is run as a container image, that will also follow.

The report is generated in a reports folder in the directory from which the cv tool is run. For example for the command:

```out/chart-verifier verify pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz```

a report is generated  ```reports/chart-0.1.0-v3.valid.tgz/verifier.report.yaml```

if the tool is re-run the existing report is overwritten. Content for the report is as follows:

```ok: true
metadata:
    tool:
        verifier-version: 1.0.0
        chart-uri: pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz
    chart:
        name: chart
        version: 1.16.0
results:
    contains-test:
        ok: true
        reason: Chart test files exist
    contains-values:
        ok: true
        reason: Values file exist
    contains-values-schema:
        ok: true
        reason: Values schema file exist
    has-minkubeversion:
        ok: true
        reason: Minimum Kubernetes version specified
    has-readme:
        ok: true
        reason: Chart has a README
    helm-lint:
        ok: true
        reason: Helm lint successful
    is-helm-v3:
        ok: true
        reason: API version is V2, used in Helm 3
    not-contain-csi-objects:
        ok: true
        reason: CSI objects do not exist
    not-contains-crds:
        ok: true
        reason: Chart does not contain CRDs
chart-yaml:
    apiVersion: v2
    appVersion: 1.16.0
    description: A Helm chart for Kubernetes
    icon: https://www.example.com/chart-icon.png
    kubeVersion: 1.20.0
    name: chart
    type: application
    version: 0.1.0-v3.valid```

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -28,6 +28,10 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
+const (
+	Version = "1.0.0"
+)
+
 func init() {
 	allChecks = chartverifier.DefaultRegistry().AllChecks()
 }
@@ -96,6 +100,7 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 				SetChecks(checks).
 				SetConfig(config).
 				SetOverrides(setOverridesFlag).
+				SetToolVersion(Version).
 				Build()
 
 			if err != nil {
@@ -124,6 +129,17 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 				cmd.Println(string(b))
 			} else {
 				cmd.Print(result)
+			}
+
+			reportErr := chartverifier.
+				NewReportBuilder().
+				SetCertificate(&result).
+				SetChartUri(args[0]).
+				Generate()
+
+			if reportErr != nil {
+				cmd.Println("Report failure :" + reportErr.Error())
+				return err
 			}
 
 			return nil
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -28,9 +28,7 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
-const (
-	Version = "1.0.0"
-)
+var Version = "1.0.0"
 
 func init() {
 	allChecks = chartverifier.DefaultRegistry().AllChecks()
--- a/pkg/chartverifier/reportbuilder.go
+++ b/pkg/chartverifier/reportbuilder.go
@@ -1,0 +1,105 @@
+package chartverifier
+
+import (
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
+	"gopkg.in/yaml.v3"
+	"helm.sh/helm/v3/pkg/chart"
+	"os"
+	"path/filepath"
+)
+
+type ReportBuilder interface {
+	SetCertificate(*Certificate) ReportBuilder
+	SetChartUri(string) ReportBuilder
+	AddChartYaml(*chart.File) ReportBuilder
+	Generate() error
+}
+
+type reportBuilder struct {
+	Certificate *Certificate
+	ChartUri    string
+	ChartYaml   *chart.File
+}
+
+type chartYamlMetadata struct {
+	ChartYamlMetadata interface{} `json:"chart-yaml" yaml:"chart-yaml"`
+}
+
+func newChartYamlMetadata(data interface{}) *chartYamlMetadata {
+	return &chartYamlMetadata{
+		ChartYamlMetadata: data,
+	}
+}
+
+func NewReportBuilder() ReportBuilder {
+	return &reportBuilder{}
+}
+
+func (r *reportBuilder) SetCertificate(cert *Certificate) ReportBuilder {
+	r.Certificate = cert
+	return r
+}
+
+func (r *reportBuilder) SetChartUri(uri string) ReportBuilder {
+	r.ChartUri = uri
+	return r
+}
+
+func (r *reportBuilder) AddChartYaml(file *chart.File) ReportBuilder {
+	r.ChartYaml = file
+	return r
+}
+
+func (r *reportBuilder) Generate() error {
+
+	var err error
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by mmulholla** [comment]
Also note some of the extra data added to the report also appears in console output, for example:

```mmulholl-mac:chart-verifier martinmulholland$ out/chart-verifier verify pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz 
Tool:
  verifier-version: 1.0.0
  chart-uri: pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz
Chart:
  Name: chart
  version: 1.16.0
ok: true

not-contains-crds:
	ok: true
	reason: Chart does not contain CRDs
has-minkubeversion:
	ok: true
	reason: Minimum Kubernetes version specified
contains-test:
	ok: true
	reason: Chart test files exist
helm-lint:
	ok: true
	reason: Helm lint successful
is-helm-v3:
	ok: true
	reason: API version is V2, used in Helm 3
not-contain-csi-objects:
	ok: true
	reason: CSI objects do not exist
has-readme:
	ok: true
	reason: Chart has a README
contains-values:
	ok: true
	reason: Values file exist
contains-values-schema:
	ok: true
	reason: Values schema file exist ```

**Comment by mmulholla** [comment]
Have broken some test based on the output of the tool, working to fix them.

**Comment by mmulholla** [comment]
> Have broken some test based on the output of the tool, working to fix them.

Commit added to fix the tests

File: pkg/chartverifier/reportbuilder.go [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+	var err error
+	reportsDir := "reports"
+	if _, err = os.Stat(reportsDir); os.IsNotExist(err) {
+		err = os.Mkdir(reportsDir, 0755)
+	}
+	if err != nil {
+		return err
+	}
+
+	reportDir := filepath.FromSlash(reportsDir + "/" + filepath.Base(r.ChartUri))
```
baijum, line 64: Please use https://golang.org/pkg/path/filepath/#Join

pkg/chartverifier/reportbuilder.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if _, err = os.Stat(reportDir); !os.IsNotExist(err) {
+		os.RemoveAll(reportDir)
+	}
+	if err = os.Mkdir(reportDir, 0755); err != nil {
+		return err
+	}
+
+	f, err := os.Create(reportDir + "/verifier.report.yaml")
```
baijum, line 73: Please use https://golang.org/pkg/path/filepath/#Join

pkg/chartverifier/reportbuilder.go
```diff
@@ -0,0 +1,105 @@
...
+
+	b, err := yaml.Marshal(r.Certificate)
+	if err != nil {
+		return err
+	}
+
+	f.Write(b)
+
+	c, _, err := checks.LoadChartFromURI(r.ChartUri)
+	for _, chartFile := range c.Raw {
```
baijum, line 86: We can use https://pkg.go.dev/helm.sh/helm/v3@v3.5.2/pkg/chart#Metadata
It is available as `c.Metadata` object here.

baijum (changes requested):

pkg/chartverifier/reportbuilder.go
mmulholla: OK

pkg/chartverifier/reportbuilder.go
mmulholla: OK

pkg/chartverifier/reportbuilder.go
mmulholla: c.Metadata contains more the the chart.yaml

```c. Metadata:
    name: chart
    home: ""
    sources: []
    version: 0.1.0-v3.valid
    description: A Helm chart for Kubernetes
    keywords: []
    maintainers: []
    icon: https://www.example.com/chart-icon.png
    apiversion: v2
    condition: ""
    tags: ""
    appversion: 1.16.0
    deprecated: false
    annotations: {}
    kubeversion: 1.20.0
    dependencies: []
    type: application
chart-yaml:
    apiVersion: v2
    appVersion: 1.16.0
    description: A Helm chart for Kubernetes
    icon: https://www.example.com/chart-icon.png
    kubeVersion: 1.20.0
    name: chart
    type: application
    version: 0.1.0-v3.valid```

Comment (baijum) [reply pkg/chartverifier/reportbuilder.go:86]: Let's use `c.Metadata `. I think that would be the correct approach.

Comment (mmulholla) [reply pkg/chartverifier/reportbuilder.go]: Updated to use c.Metadata. and change the section title:
```
chart-metadata:
    name: chart
    home: ""
    sources: []
    version: 0.1.0-v3.valid
    description: A Helm chart for Kubernetes
    keywords: []
    maintainers: []
    icon: https://www.example.com/chart-icon.png
    apiversion: v2
    condition: ""
    tags: ""
    appversion: 1.16.0
    deprecated: false
    annotations: {}
    kubeversion: 1.20.0
    dependencies: []
    type: application
```

File: cmd/verify.go [hunk 4]
```diff
@@ -28,6 +28,10 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
+const (
+	Version = "1.0.0"
+)
```
baijum, line 33: One last comment!
I think we can override the version number when building the program.  No need to change this file every time.  To support that, let's use `var` instead of `const` for `Version`. This way we can change the Version number during build time like this:
```
go build -ldflags="-X 'github.com/redhat-certification/chart-verifier/cmd.Version=1.2.0'"
```

mmulholla: Done.

baijum (approved):
LGTM

PR #74, opened 2021-04-14
Only fail lint for higher than warning messages

mmulholla:
check now only fails if the highest severity message is greater than warning.

If there is an error message the reason includes all messages.

Also created some test charts which result in various message severities from lint for testing.


isutton (approved):
LGTM

PR #75, opened 2021-04-15
New certificate schema, mandatory type, chart sha and timestamp

mmulholla:
This PR covers several things

1. New schema implemented. See "Certificate" struct in chartverifier/Certificate.go 
2. Added ability to mark checks as Mandatory, Optional or Experimental. All current checks are mandatory
3. Generating a sha256 value for the chart. Basically sort all files into alphabetic order and then create sha from their content. Test case repeatedly creates sha calues for test charts making sure they are always the same.
4. Report is output to console and no longer a file. Running locallly use "command 2> report.yaml", running docker image just "command > report.yaml"
5. Default output is now yaml. 
6. Some tests had to be updated for the new changes and schema

### DIFF
--- a/pkg/chartverifier/certificate.go
+++ b/pkg/chartverifier/certificate.go
@@ -16,76 +16,88 @@
 
 package chartverifier
 
-import "strconv"
+import (
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
+	helmchart "helm.sh/helm/v3/pkg/chart"
+)
 
-type chartMetadata struct {
-	Name    string `json:"name" yaml:"name"`
-	Version string `json:"version" yaml:"version"`
+var CertificateApiVersion = "v1"
+var CertificateKind = "verify-report"
+
+type OutcomeType string
+
+const (
+	MandatoryCheckType    checks.CheckType = "Mandatory"
+	OptionalCheckType     checks.CheckType = "Optional"
+	ExperimentalCheckType checks.CheckType = "Experimental"
+
+	FailOutcomeType OutcomeType = "FAIL"
+	PassOutcomeType OutcomeType = "PASS"
+)
+
+type Certificate struct {
+	Apiversion string              `json:"apiversion" yaml:"apiversion"`
+	Kind       string              `json:"kind" yaml:"kind"`
+	Metadata   CertificateMetadata `json:"metadata" yaml:"metadata"`
+	Results    []*CheckReport      `json:"results" yaml:"results"`
 }
 
-type runMetadata struct {
-	Version  string `json:"verifier-version" yaml:"verifier-version"`
-	ChartUri string `json:"chart-uri" yaml:"chart-uri"`
+type CertificateMetadata struct {
+	ToolMetadata ToolMetadata        `json:"tool" yaml:"tool"`
+	ChartData    *helmchart.Metadata `json:"chart" yaml:"chart"`
+	Overrides    string              `json: "chart-overrides" yaml:"chart-overrides"`
 }
 
-type metadata struct {
-	RunMetadata   runMetadata   `json:"tool" yaml:"tool"`
-	ChartMetadata chartMetadata `json:"chart" yaml:"chart"`
+type ToolMetadata struct {
+	Version                    string `json:"verifier-version" yaml:"verifier-version"`
+	ChartUri                   string `json:"chart-uri" yaml:"chart-uri"`
+	Digest                     string `json:"digest" yaml:"digest"`
+	LastCertifiedTime          string `json:"lastCertifiedTime" yaml:"lastCertifiedTime"`
+	CertifiedOpenShiftVersions string `json:"certifiedOpenShiftVersions" yaml:"certifiedOpenShiftVersions"`
 }
 
-func newMetadata(name, version, chartUri, toolVersion string) *metadata {
-	return &metadata{
-		RunMetadata: runMetadata{
-			ChartUri: chartUri,
-			Version:  toolVersion,
-		},
-		ChartMetadata: chartMetadata{
-			Name:    name,
-			Version: version,
-		},
+type CheckReport struct {
+	Check   string           `json:"check" yaml:"check"`
+	Type    checks.CheckType `json:"type" yaml:"type"`
+	Outcome OutcomeType      `json:"outcome" yaml:"outcome"`
+	Reason  string           `json:"reason" yaml:"reason"`
+}
+
+func newCertificate() Certificate {
+
+	certificate := Certificate{Apiversion: CertificateApiVersion, Kind: CertificateKind}
+	certificate.Metadata = CertificateMetadata{}
+	certificate.Metadata.ToolMetadata = ToolMetadata{}
+
+	return certificate
+}
+
+func (c *Certificate) AddCheck(checkName string, checkType checks.CheckType) *CheckReport {
+	newCheck := CheckReport{}
+	newCheck.Check = checkName
+	newCheck.Type = checkType
+	newCheck.Outcome = PassOutcomeType
+	c.Results = append(c.Results, &newCheck)
+	return &newCheck
+}
+
+func (cr *CheckReport) SetResult(outcome bool, reason string) {
+	if outcome {
+		cr.Outcome = PassOutcomeType
+	} else {
+		cr.Outcome = FailOutcomeType
 	}
+	cr.Reason = reason
 }
 
-type certificate struct {
-	Ok             bool           `json:"ok" yaml:"ok"`
-	Metadata       *metadata      `json:"metadata" yaml:"metadata"`
-	CheckResultMap checkResultMap `json:"results" yaml:"results"`
-}
+func (c *Certificate) IsOk() bool {
 
-type checkResultMap map[string]checkResult
-
-type checkResult struct {
-	Ok     bool   `json:"ok" yaml:"ok"`
-	Reason string `json:"reason" yaml:"reason"`
-}
-
-func newCertificate(name, version, chartUri, toolVersion string, ok bool, resultMap checkResultMap) Certificate {
-	return &certificate{
-		Metadata:       newMetadata(name, version, chartUri, toolVersion),
-		Ok:             ok,
-		CheckResultMap: resultMap,
+	outcome := true
+	for _, check := range c.Results {
+		if check.Outcome == FailOutcomeType {
+			outcome = false
+			break
+		}
 	}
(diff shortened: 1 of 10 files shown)


pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
 
 package chartverifier
 
-import "strconv"
+import (
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
+	helmchart "helm.sh/helm/v3/pkg/chart"
```
pedjak, line 21: is it really needed to introduce package alias? IMHO `chart.Foo` reads perfectly well.

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+)
 
-type chartMetadata struct {
-	Name    string `json:"name" yaml:"name"`
-	Version string `json:"version" yaml:"version"`
-}
+var CertificateApiVersion = "v1"
```
pedjak, line 24: is this not actually `ReportApiVersion`? The tool does not generate certificates, it runs checks and report them, IMHO? BTW, I would extract Report type and corresponding types to a separate package, i.e. `pkg/chartverifier/api`?

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	if outcome {
+		cr.Outcome = PassOutcomeType
+	} else {
+		cr.Outcome = FailOutcomeType
 	}
+	cr.Reason = reason
+}
 
-	return report
+func (c *Certificate) IsOk() bool {
```
pedjak, line 93: did not we decide that the tool is not going to aggregate/reason about the checks, i.e. what are mandatory and what not?

pkg/chartverifier/certificatebuilder.go
```diff
@@ -37,62 +39,89 @@ type CheckResult struct {
...
-		return nil, errors.New("chart name must be set")
-	}
+type fileSorter struct {
+	files []*helmchart.File
+	by    func(p1, p2 *helmchart.File) bool // Closure used in the Less method.
+}
 
-	if r.ChartVersion == "" {
-		return nil, errors.New("chart version must be set")
+func (by By) Sort(files []*helmchart.File) {
```
pedjak, line 90: can it be private?

pkg/chartverifier/checks/registry.go
```diff
@@ -42,21 +42,28 @@ func (r *Result) SetResult(outcome bool, reason string) Result {
...
+	Func CheckFunc
+}
+
 type CheckFunc func(uri string, config *viper.Viper) (Result, error)
 
 type Registry interface {
-	Get(name string) (CheckFunc, bool)
-	Add(name string, checkFunc CheckFunc) Registry
+	Get(name string) (Check, bool)
+	Add(name string, check Check) Registry
```
pedjak, line 62: could `Check` struct have `name` field, so that `Add` can have only single param,  e.g. `Add(check Check)`?

pkg/chartverifier/helmcertifier.go
```diff
@@ -31,9 +31,5 @@ type CertifierBuilder interface {
 }
 
 type Certifier interface {
-	Certify(uri string) (Certificate, error)
-}
-
-type Certificate interface {
-	IsOk() bool
+	Certify(uri string) (*Certificate, error)
```
pedjak, line 34: it looks strange to me that `Certificate` can certify something. I would expect that a certificate is result of some done certification. BTW, I would stop calling this certificate and start using Report or something else.

pedjak (changes requested):

isutton (changes requested):

pkg/chartverifier/certificate.go
mmulholla: I find "chart" too general, as this is actually the helm chart and I think the alias makes the code more clear.

mmulholla:
@isutton @pedjak Some of the comments relate to use of certificate instead of report. I used certificate because it is consistent with the rest of the code. The schema is in Cerificate.go, it is created from certificatebuilder.go, the  check process is created by certifierbuilder.go and run by certifier.go. If we use Report for the schema we should change all of those, the object and funcs too. Thoughts?

pedjak:
> If we use Report for the schema we should change all of those, the object and funcs too. Thoughts?

Yes, change everything what is needed.

pkg/chartverifier/certificate.go
mmulholla: Yes, I did this as a helper function for the tests.

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	Digest                     string `json:"digest" yaml:"digest"`
+	LastCertifiedTime          string `json:"lastCertifiedTime" yaml:"lastCertifiedTime"`
+	CertifiedOpenShiftVersions string `json:"certifiedOpenShiftVersions" yaml:"certifiedOpenShiftVersions"`
 }
 
-type certificate struct {
-	Ok             bool           `json:"ok" yaml:"ok"`
-	Metadata       *metadata      `json:"metadata" yaml:"metadata"`
-	CheckResultMap checkResultMap `json:"results" yaml:"results"`
+type CheckReport struct {
```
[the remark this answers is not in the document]
mmulholla: CheckReport is a report from a check. Results is a collection of CheckReport(s), one for each check. Report covers all of the checks and the metadata.

pkg/chartverifier/certificatebuilder.go
mmulholla: Fixed

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	return certificate
 }
 
-func (c *certificate) IsOk() bool {
-	return c.Ok
+func (c *Certificate) AddCheck(checkName string, checkType checks.CheckType) *CheckReport {
+	newCheck := CheckReport{}
+	newCheck.Check = checkName
+	newCheck.Type = checkType
+	newCheck.Outcome = PassOutcomeType
```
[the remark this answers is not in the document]
mmulholla: Added.

pkg/chartverifier/certificatebuilder.go
```diff
@@ -37,62 +39,89 @@ type CheckResult struct {
...
-	CheckResultMap checkResultMap
+	Chart       *helmchart.Chart
+	Certificate Certificate
 }
 
 func NewCertificateBuilder() CertificateBuilder {
-	return &certificateBuilder{
-		CheckResultMap: checkResultMap{},
-	}
+	cB := certificateBuilder{}
```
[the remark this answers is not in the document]
mmulholla: Fixed

pkg/chartverifier/certificate.go
mmulholla: Use of certificate instead of report goes way beyond this PR. Have opened a new issue for that work: https://github.com/redhat-certification/chart-verifier/issues/78

pkg/chartverifier/certificate.go
pedjak: ok

pkg/chartverifier/helmcertifier.go
mmulholla: Use of certificate instead of report goes way beyond this PR. Have opened a new issue for that work: #78

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
+	helmchart "helm.sh/helm/v3/pkg/chart"
+)
 
-type chartMetadata struct {
-	Name    string `json:"name" yaml:"name"`
-	Version string `json:"version" yaml:"version"`
-}
+var CertificateApiVersion = "v1"
+var CertificateKind = "verify-report"
```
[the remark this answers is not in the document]
mmulholla: Use of certificate instead of report goes way beyond this PR. Have opened a new issue for that work: #78

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	ExperimentalCheckType checks.CheckType = "Experimental"
 
-type runMetadata struct {
-	Version  string `json:"verifier-version" yaml:"verifier-version"`
-	ChartUri string `json:"chart-uri" yaml:"chart-uri"`
+	FailOutcomeType OutcomeType = "FAIL"
+	PassOutcomeType OutcomeType = "PASS"
+)
+
+type Certificate struct {
```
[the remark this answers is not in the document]
mmulholla: Use of certificate instead of report goes way beyond this PR. Have opened a new issue for that work: #78

pkg/chartverifier/certificate.go
```diff
@@ -16,76 +16,88 @@
...
+	Apiversion string              `json:"apiversion" yaml:"apiversion"`
+	Kind       string              `json:"kind" yaml:"kind"`
+	Metadata   CertificateMetadata `json:"metadata" yaml:"metadata"`
+	Results    []*CheckReport      `json:"results" yaml:"results"`
 }
 
-type metadata struct {
-	RunMetadata   runMetadata   `json:"tool" yaml:"tool"`
-	ChartMetadata chartMetadata `json:"chart" yaml:"chart"`
+type CertificateMetadata struct {
```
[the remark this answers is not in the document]
mmulholla: Use of certificate instead of report goes way beyond this PR. Have opened a new issue for that work: #78

pkg/chartverifier/certificate.go
pedjak: can you move it then under `certificate_test.go` file, or at least make it private?

pkg/chartverifier/checks/registry.go
mmulholla: Added. This must be a go v Java thing :-), I still think in Java lol

mmulholla:
@pedjak @isutton  : have addressed the comments which are not connected to using report instead of certificate - for those comments and the entire rename effort I opened a new issue #78

pkg/chartverifier/certificate.go
mmulholla: I would ask to defer on this to #78 also. The method was not introduced by this PR and is the only method defined in a Certificate interface. To remove the interface creates knock on effects in other places and I can clean it up in #78.

pedjak (approved):

PR #77, opened 2021-04-19
Include Helm like values options in verify command

isutton:
This PR implements Helm like values options in the verify command, using the same 
semantics as Helm.


mmulholla:
LGTM

mmulholla (approved):
LGTM

PR #82, opened 2021-04-21
add valid images to chart-0.1.0-v3.valid.tgz

mmulholla:
Updated chart-0.1.0-v3.valid.tgz to include valid images and updated tests to reflect this.

Getting read for the demo.


pkg/chartverifier/checks/checks_test.go
```diff
@@ -369,22 +369,26 @@ func TestImageCertify(t *testing.T) {
...
-			for i := 0; i < tc.numErrors; i++ {
-				require.Contains(t, r.Reason, ImageNotCertified)
-				r.Reason = strings.Replace(r.Reason, "_replaced_", ImageNotCertified, 1)
+			if tc.numErrors == 0 {
+				require.True(t, r.Ok)
+			} else {
+				require.False(t, r.Ok)
+				for i := 0; i < tc.numErrors; i++ {
+					require.Contains(t, r.Reason, ImageNotCertified)
+					r.Reason = strings.Replace(r.Reason, "_replaced_", ImageNotCertified, 1)
```
isutton, line 390: I don't understand this `for` loop, since the `i` variable isn't used inside the block, and `r.Reason` is unlikely to have the string `_replaced_`; in the case it does exist, then in the case of multiple errors the reason will fix itself by replacing that string with the value of `ImageNotCertified`.

Could someone shed a light on this?

isutton: Also, `require.Contains` stop the current test, meaning the replacement won't be even tried, which makes me think the line can be safely removed.

mmulholla: Looks like the replace is wrong, it should be 
```strings.Replace(r.Reason, ImageNotCertified, "_replaced_",1)```

This is testing image certification and one error will be returned for each image which fails certification. NumErrors defines how many are expected. The replace should overwrite the one that is found so that the next loop does not find the same one again. Will fix the replace. Let me know if you want other changes, not sure how you loop numErrors times without i.

mmulholla: I updated the replace invocation.

isutton:
LGTM

isutton (approved):

PR #86, opened 2021-04-23
Remove output as its not helping much

baijum:
The output is also making README lengthy

* Remove `-it` switch (interactive terminal is not required for running)


mmulholla (approved):
LGTM

PR #91, opened 2021-04-26
address three issues in image certification

mmulholla:
Issue 1:
IBM image refs used single quotes: 'cp.icr.io/cp/ibm-ucv-continuous-release-poller:2.3.0.3'. Added code to remove single quotes in addition to double quotes.

Issue 2:
In the loop which processes images the err value was not cleared for each loop so an error from previous loop could affect the next loop

Issue 3: image reference parsing assumed a registry was specified if the image ref included two "/" and used the string before the first "/" as the registry. Code now requires the registry to contain a "." and accepts a registry even when an image ref includes only on "/"


pkg/chartverifier/checks/helm.go
```diff
@@ -211,7 +211,7 @@ func getImageReferences(chartUri string, vals map[string]interface{}) ([]string,
 			line := scanner.Text()
 			line = strings.ReplaceAll(line, " ", "")
 			if strings.HasPrefix(line, "image:") {
-				image := strings.Trim(strings.TrimLeft(line, "image:"), "\"")
+				image := strings.Trim(strings.TrimLeft(line, "image:"), "\"'")
```
pedjak, line 214: do we have a test covering this change?

mmulholla: No test, but I did run it against the IBM chart which showed the failure.

pedjak: Let's add a test, at least a unit one.

pedjak: The current parsing logic works on the level of plain string processing, and this is very fragile IMHO - we should take into account the fact that these resources are yaml objects and treat them so - parse them into maps for example:

https://www.ribice.ba/golang-yaml-string-map/

and look for image references.

mmulholla: If the code was fragile I would love to get some examples of how it would break in cases where processing as yaml would not, not aware of any myself. You might say that the code might be better if it processed an image reference as yaml which I would agree with and I have now implemented. Also added a test case by adding extra images to chart-0.1.0-v3.valid.tgz. To save you exploding and checking I added some images to test-connection.yaml using single, double and no quotes:

spec:
  containers:
    - name: wget
      image: 'snyk/kubernetes-operator'
      command: ['wget']
      args: ['{{ include "chart.fullname" . }}:{{ .Values.service.port }}']
    - name: web
      image: rhscl/mongodb-36-rhel7:latest
    - name: os
      image: "openshift3/ose-pod:v3.11"

pedjak: yaml spec is not an easy beast, field values could be wrapper using single/double quotes, then there are multiline values, not quotes at all. Consider this example:

https://play.golang.org/p/jyNl7wohCFZ

So instead treating each line as a separate yaml instance, how about parsing things at once, and then looking for `image` keys?

mmulholla: That is quite a complex solution for two reasons:

1. helm template does not return valid yaml, it returns a yaml snippet for each of multiple files as single string. To process this as yaml it must be first split into the snippets for each file. If you feed the entire string into unMarshall you just get the first snippet.

2. an image reference could be buried inside any number of maps and arrays so searching through the unarshalled output is not trivial.
   
Obviously this is possible but non trivial so before I proceed can you confirm:
- This is what you intended.  
- The chances of an image reference spanning multiple lines are high enough to warrant this rewrite. 

If we do proceed on this route a concern would be relying on helm template producing consistent output and need to realize that split the template output based on a line, for example:
```
---
# Source: chart/templates/deployment.yaml 
```
will require string manipulation.

pedjak: > helm template does not return valid yaml, it returns a yaml snippet for each of multiple files as single string. To process this as yaml it must be first split into the snippets for each file. If you feed the entire string into unMarshall you just get the first snippet.

It returns multiple documents, separated by `---` - this is part of [yaml specification, see example 2.7](https://yaml.org/spec/1.2/spec.html#id2760395). You are right, such content cannot be processed with `UnMarshall`, instead we should use `yaml.Decoder`, here is the code that demonstrate how to parse such content:

https://play.golang.org/p/GL3adAuagXB

> an image reference could be buried inside any number of maps and arrays so searching through the unarshalled output is not trivial.

This boils down to a classical visitor pattern (e.g. depth-first), and recursion is our friend :)

> Obviously this is possible but non trivial so before I proceed can you confirm:

We do not need to anything in this PR, if your estimate is that the change is not trivial - current change fixes the reported problem. However, in order to avoid any surprises in the future, it would be good to move us from raw parsing and start treating result of `action.RenderManifests` as yaml documents. And this can be done in  a separate PR.

pedjak (approved):

pkg/chartverifier/checks/helm.go
mmulholla: Opened and issue to update the code in the future : #93

PR #96, opened 2021-05-01
add support for image references with sha values

mmulholla:
Code has been updated to parse image references which include "@sha" after the last "/" as a reference with a sha value instead of a tag value. 

When a sha value is found it is compared with the  parsed_data.docker_image_digest from the data a returned by Pyxis. 

All tests involving images have been extended to include image references with sha values.

I also updated a couple of messages output for failures after comments by IBM.


pkg/chartverifier/pyxis/pyxis.go
```diff
@@ -92,7 +97,7 @@ func GetImageRegistries(repository string) ([]string, error) {
 	return registries, err
 }
 
-func IsImageInRegistry(repository string, version string, registry string) (bool, error) {
+func IsImageInRegistry(registry string, repository string, imageTag string, imageSha string) (bool, error) {
```
pedjak, line 100: Since we introduce in this PR new struct `ImageReference`, why do not we use it here as the function arg? For example:

```
func IsImageInRegistry(ref *ImageReference) (bool, error)
```

pkg/chartverifier/checks/checks.go
```diff
@@ -285,28 +292,34 @@ func ImagesAreCertified(opts *CheckOptions) (Result, error) {
 	return r, nil
 }
 
-func getImageParts(image string) ([]string, string, string) {
+func getImageParts(image string) ImageReference {
```
pedjak, line 295: how about to rename this function to `getImageReference` or `parseImageReference` to reflect better what it is done here?

pkg/chartverifier/checks/checks_test.go
```diff
@@ -403,28 +412,34 @@ func TestImageParsing(t *testing.T) {
 		expectedVersion  string
 		expectedRepo     string
 		expectedRegistry string
+		expectedSha      string
```
pedjak, line 415: we could replace all these with `expectedImageRef *ImageReference`?

pkg/chartverifier/pyxis/pyxis.go
```diff
@@ -114,33 +119,51 @@ func IsImageInRegistry(repository string, version string, registry string) (bool
...
-					err = errors.New(fmt.Sprintf("Version %s not found. Found : %s", version, strings.Join(tags, ", ")))
+					if len(imageSha) > 0 {
+						err = errors.New(fmt.Sprintf("Digest %s not found. Found : %s", imageSha, strings.Join(shas, ", ")))
+					} else {
+						err = errors.New(fmt.Sprintf("Tag %s not found. Found : %s", imageTag, strings.Join(tags, ", ")))
+					}
 				}
 			} else {
-				err = errors.New(fmt.Sprintf("Registry not found: %s", registry))
+				err = errors.New(fmt.Sprintf("No images found for Registry/Component: %s/%s", registry, repository))
```
pedjak, line 166: do you mean here: `No images found for Registry/Repo` ?

pedjak (commented):
overall ok - a few comments that can improve readability

pkg/chartverifier/pyxis/pyxis.go
```diff
@@ -114,33 +119,51 @@ func IsImageInRegistry(repository string, version string, registry string) (bool
...
 				found = false
 				for _, reg := range registriesBody.PyxisRegistries {
-					for _, repo := range reg.Repositories {
-						if repo.Repository == repository && repo.Registry == registry {
-							for _, tag := range repo.Tags {
-								if tag.Name == version {
-									found = true
-									break
+					if len(imageSha) > 0 {
+						if strings.Compare(imageSha, reg.ParsedData.Digest) == 0 {
```
baijum, line 128: From the docs: https://golang.org/pkg/strings/#Compare
```
Compare is included only for symmetry with package bytes.
It is usually clearer and always faster to use the built-in string
comparison operators ==, <, >, and so on.
```

pkg/chartverifier/pyxis/pyxis.go
```diff
@@ -114,33 +119,51 @@ func IsImageInRegistry(repository string, version string, registry string) (bool
...
+							found = true
+						} else {
+							shas = append(shas, reg.ParsedData.Digest)
+						}
+					} else {
+						for _, repo := range reg.Repositories {
+							if repo.Repository == repository && repo.Registry == registry {
+								if len(imageSha) > 0 {
+
 								} else {
```
baijum, line 138:
```suggestion
								if len(imageSha) == 0 {

```

pkg/chartverifier/pyxis/pyxis.go
mmulholla: I made this change but there was quite a knock on effect:

1. in ImageReference there is an array of registries. As a result had to move the loop logic from the caller of IsImageInRegistry to inside IsImageInRegistry

2. Pyxis code cannot import checks.ImageReference without creating a circular dependency so moved the structure to Pyxis.go and now checks.go uses Pyxis.ImageReference

3. The tests for  IsImageInRegistry had to be updated to provide an ImageReference instead of individual values.

pkg/chartverifier/checks/checks.go
mmulholla: Done

pkg/chartverifier/checks/checks_test.go
mmulholla: Done

pkg/chartverifier/pyxis/pyxis.go
mmulholla: Thanks for spotting that, now fixed.

pkg/chartverifier/pyxis/pyxis.go
mmulholla: OK and done.

pkg/chartverifier/pyxis/pyxis.go
mmulholla: Ooops! Done.

mmulholla:
@baijum @pedjak, thanks for the comments, all were good so I count this as a great set of reviews! Fixed everything you raised so ready for final review.

pedjak (approved):

PR #98, opened 2021-05-04
Store error from yaml in a new variable

mmulholla:
Quick bug fix. 

Code unMarshalls helm template output line by line, if it fails for a line it is ignored and the next line is processed. Problem was the error was saved in the same variable used to return an error from the enclosing function so if harmless yaml error occurs the caller thinks the function failed, and image certification fails.

Updated code to use a specific error variable for unMarshall errors.


isutton (approved):
LGTM.

PR #102, opened 2021-05-07
Change has-minkubeversion to has-kubeversion

mmulholla:
Not much more to say than the title.

Note: should not be merged until after this PR otherwise reports will start failing checks:
https://github.com/openshift-helm-charts/charts/pull/122


baijum (approved):
LGTM

isutton (approved):

PR #103, opened 2021-05-12
Add chart-testing dependencies to chart-verifier's container image

isutton:
The `chart-testing` check depends on third party binaries
(specifically `kubectl` and `helm`), so those should be included in
the resulting image.

OpenShift's `oc` should be provided as well (please note it already
contains a symbolic link to `kubectl`), which is a chart-verifier
dependency to obtain the version of the available OpenShift
connection.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,10 +19,16 @@ RUN go mod download
 
 COPY . .
 
+RUN ./hack/get-oc.sh
+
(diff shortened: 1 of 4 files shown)


baijum (approved):

dperaza4dustbit (approved):
Looks good to me

PR #106, opened 2021-05-14
chart-verifier output contains more than the report

isutton:
The chart-verifier output since the introduction of the chart-testing
check includes `helm test` and other command output that aren't
expected.

The output should contain only the report's YAML (or JSON)
representation.


baijum (approved):
LGTM

PR #115, opened 2021-05-15
Set openshiftVersion value through a CLI flag

### LINKED ISSUE #111: Chart test passing but OS Version annotation set to N/A
In this PR the chart testing passed but the OCVersion annotation was not set. Need to understand why:

https://github.com/openshift-helm-charts/charts/pull/159

baijum:
The `oc version` command gives the `openshiftVersion` value if the
logged-in user (role) has access to get values of `clusteroperators`
(a cluster scoped resource in the `config.openshift.io` API group).
A developer may not have these kinds of access to run the chart-verifier.

This solution introduces a command-line flag to set the value of
`openshiftVersion`.  The flag is `--openshift-version` (`-V` as a short
option).  If the user has access to get values of `clusteroperators`
resource, that value will get precedence over the CLI flag.

Fixes #111

- [x] Update documentation

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"encoding/json"
+
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/cli/values"
 	"helm.sh/helm/v3/pkg/getter"
@@ -49,6 +50,8 @@ var (
 	outputFormatFlag string
 	// setOverridesFlag contains the overrides the user has specified through the --set flag.
 	setOverridesFlag []string
+	// openshiftVersionFlag set the value of `certifiedOpenShiftVersions` in the report
+	openshiftVersionFlag string
 )
 
 func filterChecks(set []string, subset []string, setEnabled bool, subsetEnabled bool) ([]string, error) {
@@ -124,6 +127,7 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 				SetConfig(config).
 				SetOverrides(verifyOpts.Values).
 				SetToolVersion(Version).
+				SetOpenShiftVersion(openshiftVersionFlag).
 				Build()
 
 			if err != nil {
@@ -174,6 +178,7 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 	cmd.Flags().StringSliceVarP(&verifyOpts.Values, "set", "s", []string{}, "overrides a configuration, e.g: dummy.ok=false")
 
 	cmd.Flags().StringSliceVarP(&verifyOpts.ValueFiles, "set-values", "f", nil, "specify application and check configuration values in a YAML file or a URL (can specify multiple)")
+	cmd.Flags().StringVarP(&openshiftVersionFlag, "openshift-version", "V", "", "version of OpenShift used in the cluster")
 
 	return cmd
 }
--- a/cmd/verify_test.go
+++ b/cmd/verify_test.go
@@ -109,6 +109,7 @@ func TestCertify(t *testing.T) {
 
 		cmd.SetArgs([]string{
 			"-e", "is-helm-v3",
+			"-V", "4.9",
 			"../pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz",
 		})
 		require.NoError(t, cmd.Execute())
@@ -131,6 +132,7 @@ func TestCertify(t *testing.T) {
 
 		cmd.SetArgs([]string{
 			"-e", "is-helm-v3", // only consider a single check, perhaps more checks in the future
+			"-V", "4.9",
 			"-o", "json",
 			"../pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz",
 		})
@@ -157,6 +159,7 @@ func TestCertify(t *testing.T) {
 
 		cmd.SetArgs([]string{
 			"-e", "is-helm-v3", // only consider a single check, perhaps more checks in the future
+			"-V", "4.9",
 			"-o", "yaml",
 			"../pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz",
 		})
--- a/docs/helm-chart-checks.md
+++ b/docs/helm-chart-checks.md
@@ -75,7 +75,6 @@ This section provides help on the basic usage of Helm chart checks with the podm
   The output is similar to the following example:
   ```
   Verifies a Helm chart by checking some of its characteristics
-  Verifies a Helm chart by checking some of its characteristics
 
   Usage:
     chart-verifier verify <chart-uri> [flags]
--- a/docs/helm-chart-checks.md
+++ b/docs/helm-chart-checks.md
@@ -57,58 +57,80 @@ To perform the tasks related to Helm chart checks, run the following commands:
 
 - Runs all the available checks for the chart using a `uri`.
 
- ```ruby
- $ docker run -it --rm quay.io/redhat-certification/chart-verifier verify <chart-uri>
- ```
+  ```
+   $ docker run -it --rm quay.io/redhat-certification/chart-verifier verify <chart-uri>
+   ```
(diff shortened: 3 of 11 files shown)


pkg/chartverifier/certifier.go
```diff
@@ -67,12 +76,9 @@ func (c *certifier) Certify(uri string) (*Certificate, error) {
...
 	oc := tool.NewOc(procExec)
 
-	osVersion, err := oc.GetVersion()
+	osVersion, err := oc.GetVersion(c.openshiftVersion)
 	if err != nil {
-		// NOTE(isutton): oc or a suitable kubeconfig might not be
-		//                available, so for now we leave it as "N/A"
-		//                instead of raising an error.
-		osVersion = "N/A"
+		return nil, OpenShiftVersionErr{err.Error()}
```
isutton, line 81: Revert `oc.GetVersion()` and in case of error use `c.openshiftVersion` as fallback, and then erroring if neither is available.

pkg/tool/oc.go
```diff
@@ -19,9 +19,9 @@ func NewOc(exec exec.ProcessExecutor) Oc {
 
 const osVersionKey = "openshiftVersion"
 
-func (o Oc) GetVersion() (string, error) {
+func (o Oc) GetVersion(ov string) (string, error) {
```
isutton, line 22: This change is not required; to use the value coming from the command line as fallback just do that in the caller.

pkg/tool/oc.go
```diff
@@ -32,7 +32,10 @@ func (o Oc) GetVersion() (string, error) {
 
 	version := out[osVersionKey]
 	if version == nil {
-		return "", fmt.Errorf("%q not found in 'oc version' output", osVersionKey)
+		if ov != "" {
+			return ov, nil
+		}
+		return "", fmt.Errorf("%q not found in 'oc version' output.  And the 'openshift-version' flag has not set.", osVersionKey)
```
isutton, line 38: Perhaps

isutton (commented):
Please revert the changes in `oc.GetVersion`, since it is meant to represent what the `oc` tool has to say within a given context. If the context can't be provided, the caller should be responsible for deciding what to do.

pkg/chartverifier/certifier.go
baijum: Reverted. 03bd100976c1685bdec2a689611f61d3e2371157

pkg/tool/oc.go
baijum: Reverted. 03bd100976c1685bdec2a689611f61d3e2371157

cmd/verify.go
```diff
@@ -174,6 +178,7 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 	cmd.Flags().StringSliceVarP(&verifyOpts.Values, "set", "s", []string{}, "overrides a configuration, e.g: dummy.ok=false")
 
 	cmd.Flags().StringSliceVarP(&verifyOpts.ValueFiles, "set-values", "f", nil, "specify application and check configuration values in a YAML file or a URL (can specify multiple)")
+	cmd.Flags().StringVarP(&openshiftVersionFlag, "openshift-version", "V", "", "version of OpenShift used in the cluster")
```
isutton, line 181: `openshift-version` accepts any arbitrary string, but it should accept only semver compatible versions.

cmd/verify.go
```diff
@@ -174,6 +178,7 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
...
```
isutton, line 182: The `https://github.com/Masterminds/semver` package is already an implicit dependency, and it can be used to validate the user input (see `semver.Version`).

cmd/verify_test.go
```diff
@@ -157,6 +159,7 @@ func TestCertify(t *testing.T) {
 
 		cmd.SetArgs([]string{
 			"-e", "is-helm-v3", // only consider a single check, perhaps more checks in the future
+			"-V", "4.9",
```
isutton, line 162: The `openshift-version` is passed but not verified in the output report.

cmd/verify_test.go
```diff
@@ -157,6 +159,7 @@ func TestCertify(t *testing.T) {
...
 			"-o", "yaml",
```
isutton, line 163: Also, there isn't a failure check when the version can't be obtained nor given by the user.

pkg/chartverifier/certifier.go
```diff
@@ -36,17 +36,26 @@ func (e CheckErr) Error() string {
 	return "check error: " + string(e)
 }
 
+type OpenShiftVersionErr struct {
+	msg string
+}
```
isutton, line 41: This could be changed into `type OpenShiftVersionErr string` instead, since `msg` is the only field in the type.

pkg/chartverifier/certifier.go
```diff
@@ -36,17 +36,26 @@ func (e CheckErr) Error() string {
...
+
```
isutton, line 42: This requires the below hunk to be adapted as well, and the usage further on of `OpenShiftVersionErr{err.Error()}` to `OpenShiftVersionErr(err.Error())`.

pkg/chartverifier/certifier.go
```diff
@@ -69,10 +78,11 @@ func (c *certifier) Certify(uri string) (*Certificate, error) {
...
-		// NOTE(isutton): oc or a suitable kubeconfig might not be
-		//                available, so for now we leave it as "N/A"
-		//                instead of raising an error.
-		osVersion = "N/A"
+		if c.openshiftVersion != "" {
+			osVersion = c.openshiftVersion
+		} else {
+			return nil, OpenShiftVersionErr{err.Error()}
+		}
 	}
```
isutton, line 86: Perhaps inverting the `if/else` clause would read better, as in wrapping the `return` in the `if` clause. Another option would be coalescing the `err` and the `c.openshiftVersion` check: `if err != nil && c.openshiftVersion == "" { return nil, ... } else { osVersion = c.openshiftVersion }`.

cmd/verify_test.go
baijum: Currently, the tool exit with an error, and no report is produced. I don't think a check is necessary.

pkg/chartverifier/certifier.go
baijum: Done. bd97de524900698c5badb59f30df389f1b827d63

pkg/chartverifier/certifier.go
baijum: Done. bd97de524900698c5badb59f30df389f1b827d63

pkg/chartverifier/certifier.go
baijum: Done. bd97de524900698c5badb59f30df389f1b827d63

isutton:
Closes #111

cmd/verify.go
baijum: Done. f3ae774f1f1304b39bdc942c7ad2bc304ad3ddf3

cmd/verify.go
baijum: Done. f3ae774f1f1304b39bdc942c7ad2bc304ad3ddf3

cmd/verify_test.go
baijum: Done. f3ae774f1f1304b39bdc942c7ad2bc304ad3ddf3

docs/helm-chart-checks.md
```diff
@@ -75,23 +75,38 @@ This section provides help on the basic usage of Helm chart checks with the podm
   The output is similar to the following example:
   ```
   Verifies a Helm chart by checking some of its characteristics
+  Verifies a Helm chart by checking some of its characteristics
```
isutton, line 78: This can be removed.

isutton (approved):
There's still one line to be removed.

docs/helm-chart-checks.md
baijum: Removed 8109de456328d19c9a35f3d49687703ae16db8e1

PR #121, opened 2021-05-18
Correct the logic to derive OS version

baijum:
There are a couple of issues with the current logic:

1. In the current logic, if the `oc version` provides a valid server version, the checking for SemVer spec was still performed against the user input value.

2. In the current logic, if the `oc version` provides a valid server version, also there is a valid user input value, then the user input value was used.  It is against the expected behavior.  The `oc version` value should have priority over user input.

### DIFF
--- a/pkg/chartverifier/certifier.go
+++ b/pkg/chartverifier/certifier.go
@@ -40,7 +40,7 @@ func (e CheckErr) Error() string {
 type OpenShiftVersionErr string
 
 func (e OpenShiftVersionErr) Error() string {
-	return "Missing OpenShift version. " + string(e) + " And the 'openshift-version' flag has not set."
+	return "Missing OpenShift version. " + string(e) + ". And the 'openshift-version' flag has not set."
 }
 
 type OpenShiftSemVerErr string
@@ -71,6 +71,18 @@ func (c *certifier) subConfig(name string) *viper.Viper {
 	}
 }
 
+func getVersion(sysVersion, userVersion string, err error) (string, error) {
+	osVersion := sysVersion
+	if err != nil {
+		if userVersion == "" {
+			return "", err
+		} else {
+			osVersion = userVersion
+		}
+	}
+	return osVersion, nil
+}
+
 func (c *certifier) Certify(uri string) (*Certificate, error) {
 
 	chrt, _, err := checks.LoadChartFromURI(uri)
@@ -82,12 +94,9 @@ func (c *certifier) Certify(uri string) (*Certificate, error) {
 	oc := tool.NewOc(procExec)
 
 	osVersion, err := oc.GetVersion()
+	osVersion, err = getVersion(osVersion, c.openshiftVersion, err)
 	if err != nil {
-		if c.openshiftVersion == "" {
-			return nil, OpenShiftVersionErr(err.Error())
-		} else {
-			osVersion = c.openshiftVersion
-		}
+		return nil, OpenShiftVersionErr(err.Error())
 	}
 
 	if _, err := semver.NewVersion(osVersion); err != nil {
--- a/pkg/chartverifier/certifier.go
+++ b/pkg/chartverifier/certifier.go
@@ -53,6 +53,11 @@ func NewCheckErr(err error) error {
 	return CheckErr(err.Error())
 }
 
+// Versioner provides OpenShift version
+type Versioner interface {
+	getVersion(debug bool) (string, error)
+}
+
 type certifier struct {
 	config           *viper.Viper
 	registry         checks.Registry
@@ -61,6 +66,7 @@ type certifier struct {
 	toolVersion      string
 	openshiftVersion string
 	values           map[string]interface{}
+	version          Versioner
 }
 
 func (c *certifier) subConfig(name string) *viper.Viper {
@@ -71,16 +77,16 @@ func (c *certifier) subConfig(name string) *viper.Viper {
 	}
 }
 
-func getVersion(sysVersion, userVersion string, err error) (string, error) {
-	osVersion := sysVersion
-	if err != nil {
-		if userVersion == "" {
-			return "", err
-		} else {
-			osVersion = userVersion
-		}
-	}
-	return osVersion, nil
+type version struct{}
+
+func (ver *version) getVersion(debug bool) (string, error) {
+
+	procExec := exec.NewProcessExecutor(debug)
+	oc := tool.NewOc(procExec)
+
+	// oc.GetVersion() returns an error both in case the oc command can't be executed and
+	// the value for the OpenShift version key not present.
+	return oc.GetVersion()
 }
 
 func (c *certifier) Certify(uri string) (*Certificate, error) {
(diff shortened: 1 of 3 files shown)


pedjak:
can you provide in the commit message/PR description details how the current logic is problematic and what this PR changes?

pkg/chartverifier/certifier.go
```diff
@@ -82,13 +82,17 @@ func (c *certifier) Certify(uri string) (*Certificate, error) {
...
+	if err != nil {
+		if c.openshiftVersion == "" {
+			return nil, OpenShiftVersionErr(err.Error())
+		} else {
+			osVersion = c.openshiftVersion
+		}
 	}
-	if _, err := semver.NewVersion(c.openshiftVersion); err != nil {
+
+	if _, err := semver.NewVersion(osVersion); err != nil {
```
isutton, line 93: This change is good, although I believe the assignment to `osVersion` below should be performed before this branch; it also makes the change regarding the err and openshiftVersion predicate not necessary.

isutton (commented):
Assuming the defect is that `c.openshiftVersion` might be empty at the time its value is being used in `semver.NewVersion`, I believe the assignment to `osVersion` should happen before the semantic version validation takes place.

baijum:
> can you provide in the commit message/PR description details how the current logic is problematic and what this PR changes?

Updated description.

pkg/chartverifier/certifier.go
baijum: @isutton Since the logic is a bit difficult to explain. I have extracted a function and written few unit tests. Please take a look.

pkg/chartverifier/certifier.go
```diff
@@ -82,13 +94,14 @@ func (c *certifier) Certify(uri string) (*Certificate, error) {
 	oc := tool.NewOc(procExec)
 
 	osVersion, err := oc.GetVersion()
-	if err != nil && c.openshiftVersion == "" {
+	osVersion, err = getVersion(osVersion, c.openshiftVersion, err)
+	if err != nil {
```
isutton, line 98: Is it required to add a function to perform this computation? I believe what it is currently encoded is equivalent to the following:
```
if err != nil {
        return nil, err
}
if osVersion == "" && c.openshiftVersion == "" {
        return nil, OpenShiftVersionErr(err.Error())
}
if osVersion == "" {
        osVersion = c.openshiftVersion
}
```

pkg/chartverifier/certifier_test.go
```diff
@@ -111,3 +111,39 @@ func TestCertifier_Certify(t *testing.T) {
...
 	cancel()
 }
+
+func TestGetVersion(t *testing.T) {
+	t.Run("oc version error and wrong user input", func(t *testing.T) {
+		errIn := errors.New("error")
+		v, err := getVersion("", "NaN", errIn)
+		require.NoError(t, err)
+		require.Equal(t, "NaN", v)
+	})
```
isutton, line 121: Assuming the function gets inlined as reviewed, those tests could be moved to the verify command tests instead, and exercized with the equivalent command line argument.

pedjak:
> Updated description.

Where? I still see "Correct the logic to derive OS version" on the top of this page. Also, now there are 3 commits on this branch, please squash them.

isutton:
> please squash them.

Can't we "Squash and merge" at the end of the PR session?

pkg/chartverifier/certifier.go
baijum: @isutton That logic is not correct. The first condition checks for error and returns the error to the calling function.  That's ignoring the user input value of `openshiftVersion`.

baijum:
On Tue, May 18, 2021 at 6:00 PM Predrag Knezevic ***@***.***>
wrote:

> Updated description.
>
> Where? I still see "Correct the logic to derive OS version" on the top of
> this page.
>

That's the PR title.

This is the current description:

There are a couple of issues with the current logic:

   1.

   In the current logic, if the oc version provides a valid server version,
   the checking for SemVer spec was still performed against the user input
   value.
   2.

   In the current logic, if the oc version provides a valid server version,
   also there is a valid user input value, then the user input value was used.
   It is against the expected behavior. The oc version value should have
   priority over user input.

pkg/chartverifier/certifier.go
mmulholla: I took a look at the logic and Baiju's logic looks right to me and is not equivalent to above login

1. oc.GetVersion() fails and flag is not set: above returns nil,error  Baiju return nil,OpenShiftVersionErr
2. oc.getVersion() fails and flag is set correctly: above return nil,error, Baiju flag,nil 
3. oc.getVersion() fails and flag is set badly, above return nil,error, Baiju nil,OpenShiftSemVerErr

I can't fault Baiju's logic. Wish single exit was used because it makes code easier to read for me but I may be alone there.

pkg/chartverifier/certifier_test.go
mmulholla: I don't object to the function and the more tests we have the better, with the function we can have the both sets of tests.

Would rather the function also had the logic to set  OpenShiftVersionErr and OpenShiftSemVerErr to encapsulate that functionality better, and get more from the test but not enough to hold up the fix.

LGTM

mmulholla: Would like to see the test cover all scenarios, not just one.

pkg/chartverifier/certifier.go
isutton: I believe the original intent was along the following code:

``` go
// oc.GetVersion() returns an error both in case the oc command can't be executed and
// the value for the OpenShift version key not present.
osVersion, err := oc.GetVersion()

// From this point on, an error is set and osVersion is empty.
if err != nil {
        if c.openshiftVersion != "" {
                osVersion = c.openshiftVersion
        } else {
                return nil, OpenShiftVersionErr("...")
        }
}

// osVersion is guaranteed to have a value, not yet validated as a semver value.
if _, err = semver.NewVersion(osVersion); err != nil {
        return nil, OpenShiftSemVerError(err)
}

// osVersion is guaranteed to a valid value from here onwards.
...
```

To the best of my knowledge it is equivalent to the following:

``` go
// oc.GetVersion() returns an error both in case the oc command can't be executed and
// the value for the OpenShift version key not present.
osVersion, getVersionErr := oc.GetVersion()

// From this point on, an error is set and osVersion is empty.
if getVersionErr != nil && if c.openshiftVersion != "" {
        osVersion = c.openshiftVersion
} 

// osVersion is empty only if an error happened and a default value 
// informed by the user hasn't been informed.
if osVersion == "" {
        return nil, OpenShiftVersionErr(getVersionErr)
}

// osVersion is guaranteed to have a value, not yet validated as a 
// semver value.
if _, err := semver.NewVersion(osVersion); err != nil {
        return nil, OpenShiftSemVerError(err)
}
// osVersion is guaranteed to a valid value from here onwards.
...
```

Which is, to my understanding, not generic enough that we'll make use of this anywhere else, which makes me consider that adding a function for this particular logic more confusing.

Another point is that the tests being executed on this function should be performed against the certifier type (tests can be found in `pkg/chartverifier/certifier_test.go`).

Below there's an example of a test:

```go
	t.Run("Result should be positive if check exists and returns positive", func(t *testing.T) {
		c := &certifier{
			settings:         cli.New(),
			config:           viper.New(),
			registry:         checks.NewRegistry().Add(checks.Check{Name: dummyCheckName, Type: MandatoryCheckType, Func: positiveCheck}),
			requiredChecks:   []string{dummyCheckName},
			openshiftVersion: "4.9",
		}

		r, err := c.Certify(validChartUri)
		require.NoError(t, err)
		require.NotNil(t, r)
		require.True(t, r.IsOk())
	})
```

The result of such a test (the `r` variable above) should contain a valid report with the given registry, and by the fact that charttesting and oc version handling are separated, this test could be adapted to check for an error (in the case "openshiftVersion" is empty).

baijum: @isutton 
I need some more clarity on testing. Irrespective of "openshiftVersion" is empty or not, the `oc version` command gets executed and returns some value based on the user role. Do you want integration testing with the OpenShift cluster? Or pure unit test? If it's a unit test, how can I mock the external command execution?

mmulholla: The problem is that in the go test oc.Version() will always fail so we never test the case were it does not, which is exactly why we need this PR - that path was broken. Adding the function which at least means we can do some level of testing by mimicking oc.Version() success, can only be a good thing and make the function wortwhile. It does mean the code itself is maybe not exactly how we would like it be but it is a comprise. Long term we need to solve the problem of chart-testing in go unit tests. When we have that done and we have have the nice clean logic you show above.

mmulholla: I propose we keep, the function, modify the function to do this and extend the test accordingly:
```
// From this point on, an error is set and osVersion is empty.
if getVersionErr != nil && if c.openshiftVersion != "" {
        osVersion = c.openshiftVersion
} 

// osVersion is empty only if an error happened and a default value 
// informed by the user hasn't been informed.
if osVersion == "" {
        return nil, OpenShiftVersionErr(getVersionErr)
}

// osVersion is guaranteed to have a value, not yet validated as a 
// semver value.
if _, err := semver.NewVersion(osVersion); err != nil {
        return nil, OpenShiftSemVerError(err)
}
```

This way we keep the test which is a good thing and the function is a compromise so we can keep the good thing.

baijum: @isutton @mmulholla I have implemented Igor's suggestion with dependency for unit testing.

Let me know if this is fine. I am fine with function-based approach also.

mmulholla:
Once this is merged I will open an issue to remove the new function once we have better tests in place such that oc.Version() can return a valid version.

mmulholla (commented):
LGTM

mmulholla (approved):

PR #122, opened 2021-05-18
Provide script to simplify running of podman/docker commands

isutton:
As part of release publish a script which enables a partner to run the
chart verifier docker image without having to use the full
podman/docker command.

The script provided in `hack/chart-verifier.sh` manages either Podman
or Docker to execute the chart-verifier's `verify` command inside a
running container; it merely acts as a front-end for the container
image accepting initially a subset of the commands guaranteed to work
on this fashion.

The following usages are expected:

``` shell
$ ./hack/chart-verifier.sh -V 4.7 /tmp/chart/chart-0.1.0.tgz
$ ./hack/chart-verifier.sh -V 4.7 ../../tmp/chart-0.1.0.tgz
$ ./hack/chart-verifier.sh -V 4.7 file:///tmp/chart-0.1.0.tgz
$ ./hack/chart-verifier.sh -V 4.7 https://charts.example.com/chart-0.1.0.tgz
```

The configuration file pointed by the `KUBECONFIG` environment
variable is made available to the container to execute checks that
require an OpenShift cluster to be executed, and the chart file (in
the case the file path starts with `file://`, or is an absolute or
relative path) is mounted inside the `/charts/` directory inside the
container.

It is worth to mention only the `-V|--openshift-version` option has
been enabled in the script, as the script is meant to be a front-end
only for the `verify` command.

This change list contains a test written in Go that exercises the
script, and requires the `CHART_VERIFIER_ENABLE_CLUSTER_TESTING`
environment variable to be set to a truthful value for it to be
executed.


mmulholla:
Couple of things:
- scripts us main image, for release they should pull latest image. 
- scripts should check the uri specified and if it a local file run the -v $(pwd):/charts version

isutton:
@mmulholla

I'm wondering whether this shim, which requires the user to clone the
repository and find the script to be used, instead of downloading the
binary for the user's appropriate platform and have available
everything they need.

I believe that container images are provided for users to automation
tasks, and the binary as a tool in the user space, so I'm in favor of
distributing both.

What do you think?

hack/chart-verifier-docker.sh
```diff
@@ -0,0 +1,19 @@
...
+# KUBECONFIG is either the value the user has configured in the
+# environment or the default location `kubectl` uses.
+KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
+
+# CHART_VERIFIER_IMAGE is the location of the image to use to verify
+# the chart.
+CHART_VERIFIER_IMAGE="${CHART_VERIFIER_IMAGE:-quay.io/redhat-certification/chart-verifier:main}"
+
+# DOCKER contains the path for the `podman` program.
+DOCKER="${DOCKER:-/usr/bin/docker}"
```
pedjak, line 12: let's try to detect if `docker` is available on PATH and use it, instead of expecting on a certain location, for example:

```
for cr in podman docker; do
  [ -z "$CONTAINER_RUNTIME" ] && $cr --help >/dev/null 2>&1 && CONTAINER_RUNTIME=$cr
done;

```

hack/chart-verifier-podman.sh
```diff
@@ -0,0 +1,24 @@
...
+# PODMAN contains the path for the `podman` program.
+PODMAN="${PODMAN:-/usr/bin/podman}"
+
+"${PODMAN}" run --rm -i                      \
+            --log-level="error"              \
+            --systemd="false"                \
+            --cgroup-manager="cgroupfs"      \
+            --cgroups="disabled"             \
+            --events-backend="none"          \
+            --log-driver="none"              \
```
pedjak, line 20: are these options really needed? I tried it without them and it worked.

hack/chart-verifier-docker.sh
```diff
@@ -0,0 +1,19 @@
...
+
+"${DOCKER}" run --rm -i                      \
```
pedjak, line 14: with the detection above we can collapse these two scripts into single one:

```
$CONTAINER_RUNTIME run --rm -e KUBECONFIG=/.kube/config \
  -v "${KUBECONFIG}":/.kube/config \
  "${CHART_VERIFIER_IMAGE}"  "$@"
```

pedjak: Since we have just single cmd `verify`, we can even append it by default:

```
$CONTAINER_RUNTIME run --rm -e KUBECONFIG=/.kube/config \
  -v "${KUBECONFIG}":/.kube/config \
  "${CHART_VERIFIER_IMAGE}"  verify "$@"
```

so that the script can be called like `chart-verifier.sh https://foo/chart.tgz`

hack/chart-verifier-podman.sh
isutton: Since I'm testing in a more restrictive environment (Ubuntu 20.04 on WSL2) those are required to run `podman`, which are assumptions Docker doesn't have (for example, that the host system has "journald" active.

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# Execute the command.
+$CONTAINER_RUNTIME                   \
+    run --rm -i                      \
+    -e KUBECONFIG=/.kube/config      \
+    -e PATH=/usr/local/bin           \
+    $EXTRA_ARGS                      \
+    -v "${KUBECONFIG}":/.kube/config \
+    "${CHART_VERIFIER_IMAGE}"        \
+    verify ${OPENSHIFT_VERSION} ${CHART_TO_VERIFY}
```
pedjak, line 80: why do we need to pass openshift version? can this not be discovered from the passed KUBECONFIG?

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+        ;;
+esac
+
+# Execute the command.
+$CONTAINER_RUNTIME                   \
+    run --rm -i                      \
+    -e KUBECONFIG=/.kube/config      \
+    -e PATH=/usr/local/bin           \
+    $EXTRA_ARGS                      \
+    -v "${KUBECONFIG}":/.kube/config \
```
pedjak, line 78: shall we not check if KUBECONFIG is set and fail if not?

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+
+[ -z "$1" ] && {
+    echo "Chart is missing, exiting."
+    exit 1
+}
+
+case "$1" in
+    http://*|https://*)
+        CHART_TO_VERIFY=$1
+        ;;
```
pedjak, line 60: do we need insert `break` here?

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+
+        EXTRA_ARGS="${EXTRA_ARGS} -v ${CHART_HOST_BASEDIR}/${CHART_NAME}:${CHART_GUEST_BASEDIR}/${CHART_NAME}"
+        ;;
+esac
+
+# Execute the command.
+$CONTAINER_RUNTIME                   \
+    run --rm -i                      \
+    -e KUBECONFIG=/.kube/config      \
+    -e PATH=/usr/local/bin           \
```
pedjak, line 76: cannot we set `PATH` at the time when we build the image instead at runtime?

hack/chartverifier_test.go
```diff
@@ -0,0 +1,33 @@
+package hack
```
pedjak, line 1: how about that we keep `hack` folder free from go files and move tests under `/test` folder in the project root?

pkg/chartverifier/checks/charttesting.go
```diff
@@ -139,10 +139,10 @@ func testRelease(
 	cleanupHelmTests bool,
 ) error {
 	if err := kubectl.WaitForDeployments(namespace, releaseSelector); err != nil {
-		return err
+		return fmt.Errorf("waiting for deployments: %w", err)
 	}
 	if err := helm.Test(namespace, release); err != nil {
-		return err
+		return fmt.Errorf("error testing release: %w", err)
```
pedjak, line 145: is these changes really needed for the script we submit in this PR to work?

pkg/tool/helm.go
```diff
@@ -27,9 +29,11 @@ func (h Helm) InstallWithValues(chart string, valuesFile string, namespace strin
...
 	}
 
-	if _, err := h.RunProcessAndCaptureOutput("helm", "install", release, chart, "--namespace", namespace,
-		"--wait", values, h.extraArgs); err != nil {
-		return err
+    execArgs := []string{"install", release, chart, "--namespace", namespace, "--wait"}
+    execArgs = append(execArgs, values...)
+    execArgs = append(execArgs, h.extraArgs...)
+	if _, err := h.RunProcessAndCaptureOutput("helm", execArgs); err != nil {
+        return fmt.Errorf("error installing release: %s -> %w", execArgs, err)
```
pedjak, line 36: same question as above.

isutton:
> +# Execute the command.
> +$CONTAINER_RUNTIME                   \
> +    run --rm -i                      \
> +    -e KUBECONFIG=/.kube/config      \
> +    -e PATH=/usr/local/bin           \
> +    $EXTRA_ARGS                      \
> +    -v "${KUBECONFIG}":/.kube/config \
> +    "${CHART_VERIFIER_IMAGE}"        \
> +    verify ${OPENSHIFT_VERSION} ${CHART_TO_VERIFY}
>
> why do we need to pass openshift version? can this not be discovered from the passed KUBECONFIG?

The flag is not mandatory, but if present it should be forwarded to the
container's endpoint. It will be discovered from the given `KUBECONFIG`,
but happens at a late stage.

`OPENSHIFT_VERSION` might evaluate to empty string, if no value has been
informed.

isutton:
> +case "$1" in
> +    http://*|https://*)
> +        CHART_TO_VERIFY=$1
> +        ;;
>
> do we need insert `break` here?

Why would we?

        break [n]
               Exit from within a for, while, until, or select loop.  If  n  is
               specified, break n levels.  n must be ≥ 1.  If n is greater than
               the number of enclosing loops, all enclosing loops  are  exited.
               The  return  value is 0 unless n is not greater than or equal to
               1.

According to `bash`'s documentation, `case` uses `;;` to delimit the
list of commands to execute when a pattern is matched:

       case word in [ [(] pattern [ | pattern ] ... ) list ;; ] ... esac
              A case command first expands word, and tries to match it against
              each  pattern  in turn, using the matching rules described under
              Pattern Matching below.  The word is expanded using tilde expan‐
              sion,  parameter  and  variable expansion, arithmetic expansion,
              command substitution, process substitution  and  quote  removal.
              Each pattern examined is expanded using tilde expansion, parame‐
              ter and variable expansion, arithmetic expansion,  command  sub‐
              stitution,  and  process substitution.  If the nocasematch shell
              option is enabled, the match is performed without regard to  the
              case  of alphabetic characters.  When a match is found, the cor‐
              responding list is executed.  If the ;;  operator  is  used,  no
              subsequent  matches are attempted after the first pattern match.
              Using ;& in place of ;; causes execution to  continue  with  the
              list  associated  with  the  next set of patterns.  Using ;;& in
              place of ;; causes the shell to test the next  pattern  list  in
              the statement, if any, and execute any associated list on a suc‐
              cessful match.  The exit status is zero if no  pattern  matches.
              Otherwise, it is the exit status of the last command executed in
              list.

isutton:
Predrag Knezevic ***@***.***> writes:

> +# Execute the command.
> +$CONTAINER_RUNTIME                   \
> +    run --rm -i                      \
> +    -e KUBECONFIG=/.kube/config      \
> +    -e PATH=/usr/local/bin           \
>
> cannot we set `PATH` at the time when we build the image instead at runtime?

It is possible indeed; I believe this was a left-over and can be
removed. Thanks for pointing that out.

isutton:
> +package hack
>
> how about that we keep `hack` folder free from go files and move tests under `/test` folder in the project root?

Sure. Will move it.

isutton:
>> +		return fmt.Errorf("waiting for deployments: %w", err)
>  	}
>  	if err := helm.Test(namespace, release); err != nil {
> -		return err
> +		return fmt.Errorf("error testing release: %w", err)
>
> is these changes really needed for the script we submit in this PR to work?

I'm not sure anyone else is working on improving error messages, so
while I decided to improve that a little bit to at least augment the
error message with context the failure is related.

isutton:
> +    execArgs := []string{"install", release, chart, "--namespace", namespace, "--wait"}
> +    execArgs = append(execArgs, values...)
> +    execArgs = append(execArgs, h.extraArgs...)
> +	if _, err := h.RunProcessAndCaptureOutput("helm", execArgs); err != nil {
> +        return fmt.Errorf("error installing release: %s -> %w", execArgs, err)
>
> same question as above.

I'm not sure anyone else is working on improving error messages, so I
decided to improve that a little bit to at least augment the error
message with context the failure is related, including the execution of
a third-party tool we depend on.

mmulholla:
> @mmulholla
> 
> I'm wondering whether this shim, which requires the user to clone the repository and find the script to be used, instead of downloading the binary for the user's appropriate platform and have available everything they need.
> 
> I believe that container images are provided for users to automation tasks, and the binary as a tool in the user space, so I'm in favor of distributing both.
> 
> What do you think?

The script is intended as an assist for folks running from the command line. They will have to download the script to use it and we have to be clear on that in the readme. Guess once they have it they could also use it in automations :-)

Dockerfile
```diff
@@ -12,17 +12,15 @@ FROM quay.io/redhat-certification/golang:1.15 as build
...
 COPY go.sum .
-
 RUN go mod download
 
 COPY . .
-
-RUN ./hack/get-oc.sh
-
-RUN ./hack/get-helm.sh
-
```
pedjak, line 25: this looks like no-op, because it does not matter when we download helm and oc?

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "Chart is missing, exiting."
+    exit 1
+}
+
+case "$1" in
+    http://*|https://*)
+        CHART_TO_VERIFY=$1
+        ;;
+    *)
+        CHART_TO_VERIFY=${1#file://}
```
pedjak, line 62: we should check if this chart exist and fail otherwise

hack/chart-verifier.sh
```diff
@@ -0,0 +1,80 @@
...
+        CHART_GUEST_BASEDIR="/charts"
+        CHART_HOST_BASEDIR=$(dirname $(realpath "$CHART_TO_VERIFY"))
+        CHART_NAME=$(basename "$CHART_TO_VERIFY")
+        CHART_TO_VERIFY="${CHART_GUEST_BASEDIR}/${CHART_NAME}"
+
+        EXTRA_ARGS="${EXTRA_ARGS} -v ${CHART_HOST_BASEDIR}/${CHART_NAME}:${CHART_GUEST_BASEDIR}/${CHART_NAME}"
```
pedjak, line 68: how about to mount chart directly, under the same path as on the host?

```
EXTRA_ARGS="${EXTRA_ARGS} -v  "$CHART_TO_VERIFY":"$CHART_TO_VERIFY"
```
in this way, we do not need lines 63-67

isutton:
> -
> -RUN ./hack/get-oc.sh
> -
> -RUN ./hack/get-helm.sh
> -
>
> this looks like no-op, because it does not matter when we download helm and oc?

Both binaries are downloaded and cached in the builder image, to then be
copied to the resulting image.

isutton:
> +            ;;
> +    esac
> +    shift
> +done
> +
> +[ -z "$1" ] && {
> +    echo "Chart is missing, exiting."
> +    exit 1
> +}
> +
> +case "$1" in
> +    http://*|https://*)
> +        CHART_TO_VERIFY=$1
> +        ;;
> +    *)
> +        CHART_TO_VERIFY=${1#file://}
>
> we should check if this chart exist and fail otherwise

I believe the intention is to not replicate assertions `chart-verifier`
performs in the script, but prepare an environment the program can run;
given that, unless the lack of an assertion might break the shim, it
will be delegated to `chart-verifier`.

I'll double check and see if that assertion is required for the proper
execution of the shim when implementing tests for it.

isutton:
Predrag Knezevic ***@***.***> writes:

> +    echo "Chart is missing, exiting."
> +    exit 1
> +}
> +
> +case "$1" in
> +    http://*|https://*)
> +        CHART_TO_VERIFY=$1
> +        ;;
> +    *)
> +        CHART_TO_VERIFY=${1#file://}
> +        CHART_GUEST_BASEDIR="/charts"
> +        CHART_HOST_BASEDIR=$(dirname $(realpath "$CHART_TO_VERIFY"))
> +        CHART_NAME=$(basename "$CHART_TO_VERIFY")
> +        CHART_TO_VERIFY="${CHART_GUEST_BASEDIR}/${CHART_NAME}"
> +
> +        EXTRA_ARGS="${EXTRA_ARGS} -v ${CHART_HOST_BASEDIR}/${CHART_NAME}:${CHART_GUEST_BASEDIR}/${CHART_NAME}"
>
> how about to mount chart directly, under the same path as on the host?

It could be done, but I'd rather have a stable path inside the container
instead of having a variable path depending on the host's file system.

>
> ```
> EXTRA_ARGS="${EXTRA_ARGS} -v  "$CHART_TO_VERIFY":"$CHART_TO_VERIFY"
> ```
> in this way, we do not need lines 63-67

The following are accepted as input:

- http:// or https://
- absolute path
- relative path

The relative path needs to be resolved to an absolute path to be used as
volume source, so the call to `realpath` is still required; then comes
the point of mounting the chart in an arbitrary directory inside the
container which might lead to subtle bugs as more dependencies and other
artifacts are required to be shipped with the container image.

pedjak (approved):

mmulholla:
Couple of questions/comments:

Would be good to make sure the latest docker image is pulled before executing.
Can you provide some usage instruction please.

hack/chart-verifier-docker.sh
isutton: Is this supposed to be only a shim to support automated tooling or trying to enable some features but using the same semantics?

The approach I've taken was to provide the same semantics the program has, instead of simplifying like this.

hack/chart-verifier.sh
isutton: Removed in e723cbd.

hack/chartverifier_test.go
isutton: Moved in f646ba3.

pkg/chartverifier/checks/charttesting.go
isutton: This change has been reverted.

pkg/tool/helm.go
isutton: I believe this has been changed in main.

mmulholla:
I want to put this on hold for now as we concentrate on other priorities. The problem is that the script does not work on mac.

1. getopt issues. Mac getopt doesn't work as required in Mac. You need to ```brew install gnu-getopt``` and update your path to make sure you use te new version of getopt.
2. realpath: command not found. You need to ```brew install coreutils```
3. podman issues. The script picks podman because podman is found on the path, however podman requires setting up and will not run by default, also setting it up is complicatad.

mmulholla (changes requested):
see my comment.

pedjak:
> I want to put this on hold for now as we concentrate on other priorities. The problem is that the script does not work on mac.

A lot of good stuff went into that script and it would definitely make life easier for user on Linux. Instead of not merging it and keeping it away from user, why do not we introduce additional check at the top of the script and if OSX is detected to exist with a proper message?

isutton:
> I want to put this on hold for now as we concentrate on other priorities. The problem is that the script does not work on mac.
> 
>     1. getopt issues. Mac getopt doesn't work as required. You need to `brew install gnu-getopt` and update your path to make sure you use the new version of getopt.
> 
>     2. realpath: command not found. You need to `brew install coreutils`

Those are some examples of portability problems that would change even if we would distribute binaries, as the binaries depend on third party programs to work, so I agree with @pedjak on documenting it accordingly or even check if the programs being used are available in the path, and give clear instructions such as you've mentioned.

>     3. podman issues. The script picks podman because podman is found on the path, however podman requires setting up and will not run by default, also setting it up is complicated.

It attempts to use podman or docker, giving preference for podman. If you've encountered in this regard a different behavior, it is a bug.

isutton:
> > I want to put this on hold for now as we concentrate on other priorities. The problem is that the script does not work on mac.
> 
> A lot of good stuff went into that script and it would definitely make life easier for user on Linux. Instead of not merging it and keeping it away from user, why do not we introduce additional check at the top of the script and if OSX is detected to exist with a proper message?

@mmulholla, I agree with @pedjak here.

mmulholla:
It would be possible to remove use of getopts and realpath using standard bash functions and make a portable script. Podman use could probably be skipped on Mac. 

We could merge this with linux only support but i am not sure that would be a good thing to do - guess it depends somewhat on what what plalform we expect our users to be working on, but to publish a tool which only runs on linux does not in my min appear to make us look good.

I think podman client comes with oc on mac but needs access to a linux server running podman to work. Don't think we want to require that to run the script.

mmulholla:
@dperaza4dustbit your input is needed here.

zonggen:
Clean run on Fedora34 (podman) with `./hack/chart-verifier.sh -V 4.8 ~/git/charts/charts/partners/hashicorp/vault/0.12.0/vault-0.12.0.tgz` without issue.

mmulholla:
@isutton I ran with a local tgz and get an error, local src worked well:

```
./hack/chart-verifier.sh -V 4.7 charts/samples/psql-service-0.1.8-1.tgz 

./hack/chart-verifier.sh: line 12: popd: directory stack empty
Error: Chart.yaml file is missing
Usage:
  chart-verifier verify <chart-uri> [flags]
 .......
```

I also tried and failed with:
- /charts/samples/psql-service-0.1.8-1.tgz 
- ./charts/samples/psql-service-0.1.8-1.tgz 
- file:///charts/samples/psql-service-0.1.8-1.tgz

mmulholla:
Also:
```
./hack/chart-verifier.sh 
Chart is missing, exiting.
```

```
./hack/chart-verifier.sh --help
Chart is missing, exiting.
```

isutton:
Not relevant anymore, closing.

Issue #78, opened 2021-04-19
Rename certificate to report


mmulholla:
The go code used the term certificate everywhere (file names and types) but these should be report. The rework should change all uses of certifier, certificate etc, to reporter, report etc, 

Note: remove IsOK() method from certificate.

PR #132, opened 2021-05-21
rename certificate to report and certifier to verify

mmulholla:
Change 
- certificate to report
- certifier to verifier
- moved report.IsOk() to verifier_test is_OK()
renames are to  be consistent with their functionality.

see #78


baijum (approved):
LGTM

PR #133, opened 2021-05-24
Add OpenShift version if chart-testing is enabled

### LINKED ISSUE #129: Chart verifier error: Missing OpenShift version. "openshiftVersion" not found in 'oc version' output. And the 'openshift-version' flag has not set.
if you run chart verifier after  https://github.com/redhat-certification/chart-verifier/pull/115

and you run the verifer without oc installed:

```docker run -v $(pwd):/charts --rm quay.io/redhat-certification/chart-verifier:278be83 verify /charts/chart-0.1.0-v3.valid.tgz```

you get this, both running the command line and the docker image: 
```
Error: Missing OpenShift version. "openshiftVersion" not found in 'oc version' output. And the 'openshift-version' flag has not set.
Usage:
  chart-verifier verify <chart-uri> [flags]

Flags:
  -S, --chart-set strings           set values for the chart (can specify multiple or separate values with commas: key1=val1,key2=val2)
  -G, --chart-set-file strings      set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
  -X, --chart-set-string strings    set STRING values for the chart (can specify multiple or separate values with commas: key1=val1,key2=val2)
  -F, --chart-values strings        specify values in a YAML file or a URL (can specify multiple)
      --debug                       enable verbose output
  -x, --disable strings             all checks will be enabled except the informed ones
  -e, --enable strings              only the informed checks will be enabled
  -h, --help                        help for verify
      --kube-apiserver string       the address and the port for the Kubernetes API server
      --kube-as-group stringArray   group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string         username to impersonate for the operation
      --kube-ca-file string         the certificate authority file for the Kubernetes API server connection
      --kube-context string         name of the kubeconfig context to use
      --kube-token string           bearer token used for authentication
      --kubeconfig string           path to the kubeconfig file
  -n, --namespace string            namespace scope for this request
  -V, --openshift-version string    version of OpenShift used in the cluster
  -o, --output string               the output format: default, json or yaml
      --registry-config string      path to the registry config file (default "/root/.config/helm/registry.json")
      --repository-cache string     path to the file containing cached repository indexes (default "/root/.cache/helm/repository")
      --repository-config string    path to the file containing repository names and URLs (default "/root/.config/helm/repositories.yaml")
  -s, --set strings                 overrides a configuration, e.g: dummy.ok=false
  -f, --set-values strings          specify application and check configuration values in a YAML file or a URL (can specify multiple)

Global Flags:
      --config string   config file (default is $HOME/.chart-verifier.yaml)

Missing OpenShift version. "openshiftVersion" not found in 'oc version' output. And the 'openshift-version' flag has not set.
```

### PR DESCRIPTION
**Pull request by baijum**
* Ensure the OpenShift version is available if the chart-testing
  check is enabled.

* Users should be able to run a check that doesn't require a cluster.

* If the user inputs an OpenShift version that should be a valid
  SemVer version.

Fixes #129
Jira: https://issues.redhat.com/browse/HELM-188

### DIFF
--- a/pkg/chartverifier/checks/charttesting.go
+++ b/pkg/chartverifier/checks/charttesting.go
@@ -20,7 +20,7 @@ import (
 // Versioner provides OpenShift version
 type Versioner func() (string, error)
 
-var getVersion = func() (string, error) {
+func getVersion() (string, error) {
 
 	procExec := exec.NewProcessExecutor(false)
 	oc := tool.NewOc(procExec)
--- a/pkg/chartverifier/checks/charttesting.go
+++ b/pkg/chartverifier/checks/charttesting.go
@@ -18,13 +18,9 @@ import (
 )
 
 // Versioner provides OpenShift version
-type Versioner interface {
-	getVersion() (string, error)
-}
+type Versioner func() (string, error)
 
-type version struct{}
-
-func (ver *version) getVersion() (string, error) {
+var getVersion = func() (string, error) {
 
 	procExec := exec.NewProcessExecutor(false)
 	oc := tool.NewOc(procExec)
@@ -137,7 +133,7 @@ func ChartTesting(opts *CheckOptions) (Result, error) {
 		}
 	}
 
-	if versionError := setOCVersion(opts, &version{}); versionError != nil {
+	if versionError := setOCVersion(opts.AnnotationHolder, getVersion); versionError != nil {
 		return NewResult(false, versionError.Error()), nil
 	}
 
@@ -383,31 +379,31 @@ func installAndTestChartRelease(
 	return result
 }
 
-func setOCVersion(opts *CheckOptions, versioner Versioner) error {
+func setOCVersion(holder AnnotationHolder, versioner Versioner) error {
 	// oc.GetVersion() returns an error both in case the oc command can't be executed and
 	// the value for the OpenShift version key not present.
-	osVersion, getVersionErr := versioner.getVersion()
+	osVersion, getVersionErr := versioner()
 
 	// From this point on, an error is set and osVersion is empty.
-	if getVersionErr != nil && opts.AnnotationHolder.GetCertifiedOpenShiftVersionFlag() != "" {
-		osVersion = opts.AnnotationHolder.GetCertifiedOpenShiftVersionFlag()
+	if getVersionErr != nil && holder.GetCertifiedOpenShiftVersionFlag() != "" {
+		osVersion = holder.GetCertifiedOpenShiftVersionFlag()
 	}
 
 	// osVersion is empty only if an error happened and a default value
 	// informed by the user hasn't been informed.
 	if osVersion == "" {
-		opts.AnnotationHolder.SetCertifiedOpenShiftVersion("N/A")
+		holder.SetCertifiedOpenShiftVersion("N/A")
 		return OpenShiftVersionErr(getVersionErr.Error())
 	}
 
 	// osVersion is guaranteed to have a value, not yet validated as a
 	// semver value.
 	if _, err := semver.NewVersion(osVersion); err != nil {
-		opts.AnnotationHolder.SetCertifiedOpenShiftVersion("N/A")
+		holder.SetCertifiedOpenShiftVersion("N/A")
 		return OpenShiftSemVerErr(err.Error())
 	}
 
-	opts.AnnotationHolder.SetCertifiedOpenShiftVersion(osVersion)
+	holder.SetCertifiedOpenShiftVersion(osVersion)
 
 	return nil
 }
--- a/pkg/chartverifier/checks/charttesting.go
+++ b/pkg/chartverifier/checks/charttesting.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/helm/chart-testing/v3/pkg/chart"
 	"github.com/helm/chart-testing/v3/pkg/config"
 	"github.com/helm/chart-testing/v3/pkg/exec"
@@ -15,6 +16,35 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/tool"
 	"gopkg.in/yaml.v3"
 )
+
+// Versioner provides OpenShift version
+type Versioner interface {
+	getVersion() (string, error)
+}
+
+type version struct{}
+
+func (ver *version) getVersion() (string, error) {
+
+	procExec := exec.NewProcessExecutor(false)
+	oc := tool.NewOc(procExec)
+
+	// oc.GetVersion() returns an error both in case the oc command can't be executed and
+	// the value for the OpenShift version key not present.
+	return oc.GetVersion()
+}
+
+type OpenShiftVersionErr string
+
+func (e OpenShiftVersionErr) Error() string {
+	return "Missing OpenShift version. " + string(e) + ". And the 'openshift-version' flag has not set."
+}
+
+type OpenShiftSemVerErr string
+
+func (e OpenShiftSemVerErr) Error() string {
+	return "OpenShift version is not following SemVer spec. " + string(e)
+}
 
 // buildChartTestingConfiguration computes the chart testing related
 // configuration from the given check options.
@@ -105,6 +135,10 @@ func ChartTesting(opts *CheckOptions) (Result, error) {
 		if result.Error != nil {
 			return NewResult(false, result.Error.Error()), nil
 		}
+	}
+
+	if versionError := setOCVersion(opts, &version{}); versionError != nil {
+		return NewResult(false, versionError.Error()), nil
 	}
 
 	return NewResult(true, ChartTestingSuccess), nil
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by mmulholla** [comment]
>     * Ensure the OpenShift version is available if the chart-testing
>       check is enabled.
> 
>     * Users should be able to run a check that doesn't require a cluster.
> 
>     * If the user inputs an OpenShift version that should be a valid
>       SemVer version.
> 
> 
> Fixes #129

I updated the implementation to run the ocVersion check as part of the chart-testing check so that if it fail to get the ocVersion the check will fail. The messages then come out as the reason. The check is only made after the chart testing has been a success. 

Some notes: 
- I made the setOCVersion as a function so it could be tested. Once we are set up to run chart-testing checks in go tests this function can be moved inline in the check code. 
- The interface for getVersion could also be removed once chart-testing  works in go tests.
- I don't have a an ocp cluster set up so I have tested this as part of a verifier run - the chart-testing fails tor early

**Comment by baijum** [comment]
> I updated the implementation to run the ocVersion check as part of the chart-testing check

The changes look good to me. Let's wait for a review from @isutton or @pedjak

**Comment by baijum** [comment]
@mmulholla I have implemented the #139 here.

I keep the configuration to set the version explicitly to support running against a cluster version still in beta (not released).
For example, OpenShift 4.8 is still not released, so the map doesn't exist.

**Comment by mmulholla** [comment]
LGTM. As this is a combined effort by Baiju and myself, I would like at least one other reviewer.

File: pkg/chartverifier/checks/charttesting.go [hunk 1]
```diff
@@ -4,13 +4,43 @@ import (
...
 	"github.com/helm/chart-testing/v3/pkg/config"
 	"github.com/helm/chart-testing/v3/pkg/exec"
 	"github.com/helm/chart-testing/v3/pkg/util"
 	"github.com/redhat-certification/chart-verifier/pkg/tool"
 )
 
+// Versioner provides OpenShift version
+type Versioner interface {
+	getVersion() (string, error)
+}
```
isutton, line 18: Versioner could be `type Versioner func() (string, error)` instead; this makes it easier later on when overriding it in-place.

pkg/chartverifier/checks/charttesting.go
```diff
@@ -102,6 +132,10 @@ func ChartTesting(opts *CheckOptions) (Result, error) {
 		}
 	}
 
+	if versionError := setOCVersion(opts, &version{}); versionError != nil {
+		return NewResult(false, versionError.Error()), nil
```
isutton, line 136: Instead of abusing 'opts' to hold the AnnotationHolder value, I'd augment the `Result` type with a map to hold arbitrary values collected by the checks; this would effectively allow the caller to decide what to propagate upwards instead of delegating the decision to the check. IOW, `return NewResult(true, ChartTestingsuccess, map[string]string{}{"osVersion": osVersion})`.

pkg/chartverifier/checks/charttesting.go
```diff
@@ -256,3 +290,32 @@ func installAndTestChartRelease(
 
 	return result
 }
+
+func setOCVersion(opts *CheckOptions, versioner Versioner) error {
```
isutton, line 294: `opts` is being used only to obtain an AnnotationHolder in the context, so it could be changed to an AnnotationHolder value instead. I particularly don't like this domain of annotation of a report going all the way to a check that should be agnostic.

pkg/chartverifier/verifier.go
```diff
@@ -77,64 +74,33 @@ func (c *verifier) subConfig(name string) *viper.Viper {
...
 		r, checkErr := check.Func(&checks.CheckOptions{
-			HelmEnvSettings: c.settings,
-			URI:             uri,
-			Values:          c.values,
-			ViperConfig:     c.subConfig(name),
+			HelmEnvSettings:  c.settings,
+			URI:              uri,
+			Values:           c.values,
+			ViperConfig:      c.subConfig(name),
+			AnnotationHolder: &holder,
```
isutton, line 103: Another option to adding `AnnotationHolder` as a member of `CheckOptions` is to change the `check.Func` type to get a hold of the aforementioned `map[string]string` bag.

isutton (commented):
I particularly don't like the CheckOptions type to contain values  such as the AnnotationHolder for a simple reason: tests should be  treated as separated programs that do not interfere with the rest of  the verification process. Like mentioned in another opportunity, I'd  rather expose this as a structure downwards (as in adding another  field, sibling of 'Reason') instead of this indirection. IOW, if the  OpenShift version is important mainly for the chart-testing check,  then let's remove it from the report metadata and remove this  coupling.

pkg/chartverifier/checks/charttesting.go
mmulholla: Done, My first foray into function type so hopefully I did it correctly - at least the tests pass.

pkg/chartverifier/checks/charttesting.go
mmulholla: Done.

pkg/chartverifier/verifier.go
mmulholla: Also not keen on this approach having to check for a value and then have a mapping of the key used to the annotation. Once again I think we need to figure a way to do this but I am not seeing a great solution just yet.

pkg/chartverifier/checks/charttesting.go
mmulholla: Not sure I agree this is abusing opts, maybe not a perfect fit but opts can be anything you need to pass to a function to enable it to work - it is our internal structure so we define its scope. I am not a fan of putting it in results either. I want to propose leaving as-is for now so we can create a release tomorrow and open an issue to figure out how best to do this.

> I particularly don't like the CheckOptions type to contain values such as the AnnotationHolder for a simple reason: tests should be treated as separated programs that do not interfere with the rest of the verification process. Like mentioned in another opportunity, I'd rather expose this as a structure downwards (as in adding another field, sibling of 'Reason') instead of this indirection. IOW, if the OpenShift version is important mainly for the chart-testing check, then let's remove it from the report metadata and remove this coupling.

I don't think we have broken the separation here, just consider that part of the testing check is whether we can get the OpenShift version and if we can't get it the check fails. The bad/difficult part is getting into an annotation given the structure of the code. it may have been better to have an extra check to to check we can get an OpenShift version but not at this stage. Anyway, as I say above, can we agree to leave as-is for now and open an issue to fix it after the release?

isutton:
> @@ -102,6 +132,10 @@ func ChartTesting(opts *CheckOptions) (Result, error) {
>  		}
>  	}
>  
> +	if versionError := setOCVersion(opts, &version{}); versionError != nil {
> +		return NewResult(false, versionError.Error()), nil
>
> Not sure I agree this is abusing opts, maybe not a perfect fit but opts can be anything you need to pass to a function to enable it to work - it is our internal structure so we define its scope. I am not a fan of putting it in results either. I want to propose leaving as-is for now so we can create a release tomorrow and open an issue to figure out how best to do this.

When I initially introduced the type, it was intended to be treated as a
bag of input options; similarly, the Result type was intended to contain
useful information for the caller to inspect.

I find it pretty reasonable to assume that the data being returned is
munged and then used somewhere else by the caller, keeping the
boundaries clear. IOW, the simplistic design implemented has the
assumption CheckOptions is the input and Result is the output.

>
>> I particularly don't like the CheckOptions type to contain values such as the AnnotationHolder for a simple reason: tests should be treated as separated programs that do not interfere with the rest of the verification process. Like mentioned in another opportunity, I'd rather expose this as a structure downwards (as in adding another field, sibling of 'Reason') instead of this indirection. IOW, if the OpenShift version is important mainly for the chart-testing check, then let's remove it from the report metadata and remove this coupling.
>
> I don't think we have broken the separation here, just consider that
> part of the testing check is whether we can get the OpenShift version
> and if we can't get it the check fails.

This should be another required check instead, like 'is-helm-v3'. For
example, 'has-openshift-version' check could be used to extract this
information as a required check, and since 'chart-testing' is not a
required check, this chicken-and-egg problem can be solved.

> The bad/difficult part is getting into an annotation given the
> structure of the code. it may have been better to have an extra check
> to to check we can get an OpenShift version but not at this stage.

I believe we do agree, as my reply before was written before I got here
:)

> Anyway, as I say above, can we agree to leave as-is for now and open
> an issue to fix it after the release?

Sure.

pkg/chartverifier/checks/charttesting.go
```diff
@@ -16,6 +17,31 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// Versioner provides OpenShift version
+type Versioner func() (string, error)
+
+var getVersion = func() (string, error) {
```
isutton, line 23: This could be just a function definition: `func getVersion() (string, error) {...`

pkg/chartverifier/checks/charttesting_test.go
```diff
@@ -117,3 +118,80 @@ func TestChartTesting(t *testing.T) {
 		})
 	}
 }
+
+var getVersionError = func() (string, error) {
```
isutton, line 122: `func getVersionError() (string, error) {...`

pkg/chartverifier/checks/charttesting_test.go
```diff
@@ -117,3 +118,80 @@ func TestChartTesting(t *testing.T) {
...
+	return "", errors.New("error")
+}
+
+var getVersionGood = func() (string, error) {
```
isutton, line 126: `func getVersionGood() (string, error) {...`

isutton (commented):
Apart the discussion regarding CheckOptions, I think this is good but can be simplified as stated in the comments.

pkg/chartverifier/checks/charttesting.go
mmulholla: Done

pkg/chartverifier/checks/charttesting_test.go
mmulholla: Done

pkg/chartverifier/checks/charttesting_test.go
mmulholla: Done

mmulholla:
@isutton I have opened this issue to address our concerns with the current implementation: https://github.com/redhat-certification/chart-verifier/issues/146

isutton (approved):
LGTM.

Issue #135, opened 2021-05-25
Introduce debug flag in chart-verifier


isutton:
`chart-verifier` currently slurps output messages that have been
streamed to Stdout by default in chart-testing, but we would like to
instead stream those to Stderr when the `--debug` flag is passed.

komish:
@mmulholla  I believe this can be **closed** as well. The debug flag exists in the chart-verifier today.

PR #141, opened 2021-06-02
Improve reason output when chart-testing fails

### LINKED ISSUE #114: Improve reason output when chart-testing fails
For example, current reason when no OC cluster is available is: 

Error waiting for process: exit status 1

Need to make sure error message provides a good indication of the underlying problem.

isutton:
This change list improves in the Stderr and Stdout capture of helm and
kubectl/oc when errors are triggered not only by the absence of those
programs, but also when those programs are available but return
errors.

Closes #114

### DIFF
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -40,8 +40,10 @@ func (h Helm) InstallWithValues(chart string, valuesFile string, namespace strin
 		// executing helm with args 'helm install ...': <process error>
 		// ---
 		// <stdout capture>
-		_ = stdoutCapture
-		return fmt.Errorf("executing helm with args %q: %w", strings.Join(helmArgs, " "), err)
+        if stdoutCapture == "" {
+            return fmt.Errorf("executing helm with args %q: %w", strings.Join(helmArgs, " "), err)
+        }
+        return fmt.Errorf("executing helm with args %q: %w\n---\n%s", strings.Join(helmArgs, " "), err, stdoutCapture)
 	}
 
 	return nil
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -23,7 +23,7 @@ func NewHelm(exec ProcessExecutor, extraArgs []string) Helm {
 }
 
 func toStringArray(args []interface{}) []string {
-    copy := make([]string, len(args))
+	copy := make([]string, len(args))
 	for i, a := range args {
 		copy[i] = fmt.Sprint(a)
 	}
@@ -31,7 +31,7 @@ func toStringArray(args []interface{}) []string {
 }
 
 func toInterfaceArray(args []string) []interface{} {
-    copy := make([]interface{}, len(args))
+	copy := make([]interface{}, len(args))
 	for i, a := range args {
 		copy[i] = a
 	}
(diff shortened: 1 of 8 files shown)


mmulholla:
Can we also capture and include output form helm test?

isutton:
> Can we also capture and include output from helm test?

It already should since 'Helm.Test' flows through
'Helm.RunProcessInDirAndCaptureOutput' method, but it hasn't been
explicitly tested.

#135 has been created to track the implementation of a debug flag where the captured output would be streamed to the user's Stderr, if the intention of the question is debugging purposes.

mmulholla:
> > Can we also capture and include output from helm test?
> 
> It already should since 'Helm.Test' flows through
> 'Helm.RunProcessInDirAndCaptureOutput' method, but it hasn't been
> explicitly tested.
> 
> #135 has been created to track the implementation of a debug flag
> where the captured output would be streamed to the user's Stderr, if
> the intention of the question is related to debugging purposes.

```
func (h Helm) Test(namespace string, release string) error { 
        	_, err := h.RunProcessAndCaptureOutput("helm", "test", release, "--namespace", namespace, h.extraArgs)
        	return err
 }
```

the captured output is ignored.

pkg/tool/helm.go
```diff
@@ -21,15 +24,47 @@ func NewHelm(exec exec.ProcessExecutor, extraArgs []string) Helm {
...
 	}
 
-	if _, err := h.RunProcessAndCaptureOutput("helm", "install", release, chart, "--namespace", namespace,
-		"--wait", values, h.extraArgs); err != nil {
-		return err
+	helmArgs := []string{"install", release, chart, "--namespace", namespace, "--wait"}
+	helmArgs = append(helmArgs, values...)
+	helmArgs = append(helmArgs, h.extraArgs...)
+
+	if outputCapture, err := h.RunProcessAndCaptureOutput("helm", helmArgs); err != nil {
```
dperaza4dustbit, line 57: This takes care of helm install command, what about helm test command? User would also want to see errors there right? See Test function below? No changes needed there?

isutton:
> ```
> func (h Helm) Test(namespace string, release string) error { 
>         	_, err := h.RunProcessAndCaptureOutput("helm", "test", release, "--namespace", namespace, h.extraArgs)
>         	return err
>  }
> ```
>
> the captured output is ignored.

In this case, `err` has been augmented with the captured output in the
case of failure.

If we're referring to streaming the captured output of tests to the user
Stderr for debugging purposes, then #135 has been created to track that
for us.

isutton:
>  func (h Helm) InstallWithValues(chart string, valuesFile string, namespace string, release string) error {
>  	var values []string
>  	if valuesFile != "" {
>  		values = []string{"--values", valuesFile}
>  	}
>  
> -	if _, err := h.RunProcessAndCaptureOutput("helm", "install", release, chart, "--namespace", namespace,
> -		"--wait", values, h.extraArgs); err != nil {
> -		return err
> +	helmArgs := []string{"install", release, chart, "--namespace", namespace, "--wait"}
> +	helmArgs = append(helmArgs, values...)
> +	helmArgs = append(helmArgs, h.extraArgs...)
> +
> +	if outputCapture, err := h.RunProcessAndCaptureOutput("helm", helmArgs); err != nil {
>
> This takes care of helm install command, what about helm test command? User would also want to see errors there right? See Test function below? No changes needed there?
>

Right, I see the point you and @mmulholl are making. 

I'll move the error augmentation from `InstallWithValues` to
`RunProcessInDirAndCaptureOutput`, so the other two methods will inherit
the same behavior without changing its implementation.

dperaza4dustbit (approved):

pkg/tool/helm.go
isutton: This has been fixed in d960d23, where both Helm and Kubectl types both use; Oc is covered in 
2c20705.

isutton: The machinery responsible for enriching the error with the captured output content has been moved to another type and shared among all tools per d960d23 and bfb256e.

mmulholla (approved):
LGTM

dperaza4dustbit (approved):

PR #150, opened 2021-06-07
Fix Interface conversion error on oc.go

mmulholla:


baijum:
To avoid further regression, let's add some unit tests.

mmulholla:
Test looks good. Thank you Baiju.

pkg/tool/oc.go
```diff
@@ -41,8 +41,8 @@ func (o Oc) GetVersion() (string, error) {
 	}
 	// Relying on Kubernetes version can be replaced after fixing this issue:
 	// https://bugzilla.redhat.com/show_bug.cgi?id=1850656
-	kubeServerVersion := out[osVersionKey].(map[string]string)
-	kubeVersion := kubeServerVersion["major"] + "." + kubeServerVersion["minor"]
+	kubeServerVersion := out[osVersionKey].(map[string]interface{})
+	kubeVersion := fmt.Sprintf("%v.%v", kubeServerVersion["major"], kubeServerVersion["minor"])
```
baijum, line 45: Instead of `%v` format string, I think `%s` would be more appropriate.

baijum (approved):
LGTM

pkg/tool/oc.go
mmulholla: Updated

PR #153, opened 2021-06-11
add profile support

mmulholla:
Added first pass at support for profiles. It just reads a single fixed profile.User should not see any change in behavior. Added new package pkg/chartverifier/profiles which contains 4 files:

- profile-1.0.0.yaml :  first pass profile.
- profile.go : profile schema and code to read profile-1.0.0.yaml. Also a filter function to reduce a map of checks to those defined in the profile.
- default-profile.go : a loaded schema to use if the yaml cannot be read for any reason. 
- profile-test.go : tests to make sure the yaml loads correctly, and that the filter function works. Can be expanded when support for multiple profiles is added

Note we could just code all profiles in go (instead of .yaml) as in default-profile.go

Lots of changes in pkg/chartverifier/checks/registry.go
- Added single truth (constants) for check names and check types - this had a knock on effect in lots of places
    -  where check names were previously strings the are now of type checks.Name 
- Add support for check version, and created a struct of check name and check version which can be used a key into a map of checks.

pkg/chartverifier/verifierbuilder.go init function loads the check registry - new simplified signature. Check Type is now exclusively set in the profile.

pkg/chartverifier/reportbuilder.go checks annotations specified in profile to decide what to add.

cmd/verify.go calls profiler to filter checks and create a list of checks to perform

Note; Adding version support forces cmd/verify.go to work with more than just the check name. As a result it collects a list of checks which it passes the verifierbuilder instead of a list of method names. As a result verifierbuilder does not need.

Note previous behavior:
1. verfierbuilder adds checks to registry
2. verify cmd gets list of check names from registry
3. verify command modifies list based on flags
4. verify command passes modifed list to verifeirbuilder 
5. veriferbuilder gets the checks from the registry based on the passed check names.

New behavior:
1. verfierbuilder adds checks to registry
2. verify command gets a map of checks from the registry - key is check name
3. verify command uses profile to filter the map of checks to those defined in the profile
4. verify command further modifes the map of checks based on flags 
5. verify command passes the modifed map to verifierbuilder
6. veriferbuilder gets the checks from the provided map values (does not need to call the registry)


cmd/verify.go
```diff
@@ -54,27 +57,27 @@ var (
 	openshiftVersionFlag string
 )
 
-func filterChecks(set []string, subset []string, setEnabled bool, subsetEnabled bool) ([]string, error) {
-	selected := make([]string, 0)
-	seen := map[string]bool{}
-	for _, v := range set {
-		seen[v] = setEnabled
+func filterChecks(set map[checks.CheckName]checks.Check, subset []string, setEnabled bool, subsetEnabled bool) (map[checks.CheckName]checks.Check, error) {
```
isutton, line 60: `map[checks.CheckName]checks.Check` gets repeated a lot; it likely means a type might be useful.

Perhaps the already existing registry type could be used as input and output, so the profile registry is the product of a function of the default registry and the profile description.

pkg/chartverifier/profiles/profile-default.go
```diff
@@ -0,0 +1,30 @@
+package profiles
```
isutton, line 1: I'd rename this file to `default.go`, as dashes `-` are not common in Go.

cmd/verify.go
```diff
@@ -35,13 +37,14 @@ import (
 var Version = "1.0.0"
 
 func init() {
-	allChecks = chartverifier.DefaultRegistry().AllChecks()
+	registryChecks := chartverifier.DefaultRegistry().AllChecks()
+	allChecks = profiles.GetProfile().FilterChecks(registryChecks)
```
isutton, line 41: What about filtering the checks in the command runtime instead of package initialization time?

This also gives the opportunity to extend the `verify` command with a `profile` flag, serving as input for the `profiles.GetProfile()`.

pkg/chartverifier/report.go
```diff
@@ -50,16 +46,16 @@ type ReportMetadata struct {
 }
 
 type ToolMetadata struct {
-	Version                string `json:"verifier-version" yaml:"verifier-version"`
-	ChartUri               string `json:"chart-uri" yaml:"chart-uri"`
-	Digest                 string `json:"digest" yaml:"digest"`
-	LastCertifiedTimestamp string `json:"lastCertifiedTimestamp" yaml:"lastCertifiedTimestamp"`
-
-	CertifiedOpenShiftVersions string `json:"certifiedOpenShiftVersions" yaml:"certifiedOpenShiftVersions"`
+	Version                    string `json:"verifier-version" yaml:"verifier-version"`
```
isutton, line 49: I believe it is more common for fields to follow a `snakeCase` convention instead of `kebab-case`:

```suggestion
	Version                    string `json:"verifierVersion" yaml:"verifierVersion"`
```

pkg/chartverifier/report.go
```diff
@@ -50,16 +46,16 @@ type ReportMetadata struct {
...
+	Profile                    string `json:"profile-name" yaml:"profile-name"`
+	ChartUri                   string `json:"chart-uri" yaml:"chart-uri"`
+	Digest                     string `json:"digest,omitempty" yaml:"digest,omitempty"`
```
isutton, line 52: Which cases `digest` is supposed to be empty?

cmd/verify.go
mmulholla: OK I used 
DefaultRegistry type for  map[checks.CheckId]checks.Check

and:

Added in profiles.go: type FilteredRegistry map[checks.CheckName]checks.Check
Add in verifierBuilder.go: type FilteredRegistry map[checks.CheckName]checks.Check

I know these are the same but needed to avoid circular dependencies. Could just put the typoe in registry.go but that seemed inappropriate.

pkg/chartverifier/profiles/profile-default.go
mmulholla: Renamed

cmd/verify.go
mmulholla: Good point. Moved it into the BuildChecks function but this had a roll on effect of the tests which call BuildChecks because the registry passed to it is filtered based on the profile.

pkg/chartverifier/report.go
mmulholla: I updated the one I added profile-name to profileName but not the others because they are already in the reports and the submission flow may read them (at least chart-uri) e.g.: https://github.com/openshift-helm-charts/charts/blob/6ea45d9f90b8317e3b1853ec0f36fecc273a4b29/scripts/src/chartprreview/verify-report.sh#L98

pkg/chartverifier/report.go
mmulholla: With the new profile the annotations to include are configurable, so this enables the report to omit the digest if it is not required by the profile.

mmulholla:
@isutton have responded to your comments, please let me know what you think.

pkg/chartverifier/profiles/profile.go
```diff
@@ -0,0 +1,100 @@
...
+	Type checks.CheckType `json:"type" yaml:"type"`
+}
+
+type FilteredRegistry map[checks.CheckName]checks.Check
+
+var profile *Profile
+
+func GetProfile() *Profile {
+
+	if profile != nil {
```
zonggen, line 41: Isn't `profile` always `nil` in this case?

pkg/chartverifier/profiles/profile.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+type ProfileCheck struct {
+	Name string           `json:"name" yaml:"name"`
+	Type checks.CheckType `json:"type" yaml:"type"`
+}
+
+type FilteredRegistry map[checks.CheckName]checks.Check
+
+var profile *Profile
```
zonggen, line 37: Do we really need this global variable? Since from my understanding all `profile` instance values are returned and passed across function calls.

pkg/chartverifier/profiles/profile.go
mmulholla: Good spot, in the rework I forgot to add back the code to set the profile name in the report. verifierBuilder now also calls ```profiles.GetProfile().``` With two callers I don't want to read and process the yaml file twice.

pkg/chartverifier/profiles/profile.go
mmulholla: Just use it to save the loaded profile content so we do that only once (see above)

pkg/chartverifier/profiles/profile.go
```diff
@@ -0,0 +1,100 @@
...
+
+type Profile struct {
+	Apiversion  string          `json:"apiversion" yaml:"apiversion"`
+	Kind        string          `json:"kind" yaml:"kind"`
+	Name        string          `json:"name" yaml:"name"`
+	Annotations []Annotation    `json:"annotations" yaml:"annotations"`
+	Checks      []*ProfileCheck `json:"checks" yaml:"checks"`
+}
+
+type ProfileCheck struct {
```
isutton, line 30: The `ProfileCheck` type already lives in the `profiles` package, so the `Profile` prefix can be dropped as well.

mmulholla: Done. Managed to squash the previous commit with this oneso if you look at the commit you will see other changes - oops.

pkg/chartverifier/checks/registry.go
```diff
@@ -82,31 +108,29 @@ type CheckOptions struct {
...
-	Get(name string) (Check, bool)
-	Add(check Check) Registry
-	AllChecks() []string
+	Get(id CheckId) (Check, bool)
+	Add(name CheckName, version Version, checkFunc CheckFunc) Registry
+	AllChecks() DefaultRegistry
 }
 
-type defaultRegistry map[string]Check
+type DefaultRegistry map[CheckId]Check
```
pedjak, line 116: do we really need here map and not slice? Also, would not clearer if we call this type `Checks`?

pkg/chartverifier/profiles/profile.go
```diff
@@ -0,0 +1,100 @@
...
+type Check struct {
+	Name string           `json:"name" yaml:"name"`
+	Type checks.CheckType `json:"type" yaml:"type"`
+}
+
+type FilteredRegistry map[checks.CheckName]checks.Check
+
+var profile *Profile
+
+func GetProfile() *Profile {
```
pedjak, line 39: since you are already in `profile` package, then the function name can be just `Get`, IMHO, it reads better like `profiles.Get()`

pkg/chartverifier/profiles/profile_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	testProfile := Profile{}
+
+	testProfile.Apiversion = "v1"
+	testProfile.Kind = "verifier-profile"
+	testProfile.Name = "profile-1.0.0"
+
+	testProfile.Annotations = []Annotation{DigestAnnotation, OCPVersionAnnotation, LastCertifiedTimestampAnnotation}
+
+	testProfile.Checks = []*Check{
```
pedjak, line 24: is this not the same as the result of `getDefaultProfile`? could not we call that function instead here?

pkg/chartverifier/checks/registry.go
mmulholla: I think we do, a check is identified now by the check name and the version which is encapsulated in CheckId. The name could be a single string  combination of name and version but that adds complication when processing the -x and -e  flags which only specify a check name. Also if you do a slice if we want to get a check based on its name and version we would have to loop though the slice instead of getting the value based on a key.

pkg/chartverifier/profiles/profile.go
mmulholla: Would like to defer this as the api will change in the next stage for multiple profiles and will be sure to remove the duplication.

mmulholla: Done.

pkg/chartverifier/profiles/profile_test.go
mmulholla: Done, also it highlighted a mistake in the defaultProfile :-)

zonggen (commented):
Question: Do we have plans for using the version information attached to checks? If I search for "v1.0" I saw a lot of repeated string literals. Otherwise I really like the direction this PR's going :+1:

mmulholla:
> als. Otherwise I really like the direction this PR's going 👍

Have started working in multi-profile support and and I am changing the v.1.0's in there if you don't mind waiting :-).

pkg/chartverifier/report.go
isutton: How would you solve this in the long run? Add another issue to perform the change on both systems atomically?

isutton (approved):
There are some minors such as field names using a different format than the rest of the ecosystem but are orthogonal to the issue of this PR, constructors instead of `make` in internal types, but nothing that prevents this to be merged.

zonggen (approved):
As discussed in the meetings, +1 on merging.

PR #160, opened 2021-06-25
cmd/check: add check command to parse report.yaml

zonggen:
As part of work to add profile support in chart-verifier, we need to
move the hardcoded logic from chart submission workflow to chart-verifier.

An example is the
https://github.com/openshift-helm-charts/charts/blob/main/scripts/src/chartprreview/verify-report.sh#L197
where we hardcoded mandatory checks. Here we will change to use profiles
defined within verifier to minimize disparity. The next step would be to move the `getAnnotations` logic to verifier.

Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2021 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/profiles"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	defaultProfile   = "default"
+	partnerProfile   = "partner"
+	redhatProfile    = "redhat"
+	communityProfile = "community"
+)
+
+type checkResult struct {
+	Profile     string                       `json:"profile,omitempty" yaml:"profile,omitempty"`
+	Passed      int                          `json:"passed" yaml:"passed"`
+	Failed      int                          `json:"failed" yaml:"failed"`
+	Unknown     int                          `json:"unknown" yaml:"unknown"`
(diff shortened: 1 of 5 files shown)


zonggen:
Changing to Draft as this depends on multi profile support.

cmd/check.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
+	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/profiles"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	defaultProfile   = "default"
```
isutton, line 33: The profile type deserves a type:

```go
type ProfileType string

const (
    defaultProfile ProfileType = "default"
    ...
)
```

cmd/check.go
```diff
@@ -0,0 +1,176 @@
...
+	partnerProfile   = "partner"
+	redhatProfile    = "redhat"
+	communityProfile = "community"
+)
+
+type checkResult struct {
+	Profile     string                       `json:"profile,omitempty" yaml:"profile,omitempty"`
```
isutton, line 40: If `ProfileType` is defined, then:

```go
	Profile     ProfileType                       `json:"profile,omitempty" yaml:"profile,omitempty"`
```

cmd/check.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+const (
+	defaultProfile   = "default"
+	partnerProfile   = "partner"
+	redhatProfile    = "redhat"
+	communityProfile = "community"
+)
+
+type checkResult struct {
```
isutton, line 39: I am not sure whether `checkResult` should implement the `UnmarshalJSON` method to enforce the value being a valid `ProfileType` value.

cmd/check.go
isutton, line 40: Are `default` and empty string equivalent?

I would make it required and allow it to be only one of the defined types to avoid further confusion.

cmd/check.go
```diff
@@ -0,0 +1,176 @@
...
+	Profile     string                       `json:"profile,omitempty" yaml:"profile,omitempty"`
+	Passed      int                          `json:"passed" yaml:"passed"`
+	Failed      int                          `json:"failed" yaml:"failed"`
+	Unknown     int                          `json:"unknown" yaml:"unknown"`
+	Message     []*chartverifier.CheckReport `json:"message" yaml:"message"`
+	OtherResult *checkResult                 `json:"other-result,omitempty" yaml:"other-result,omitempty"`
```
isutton, line 45:
```suggestion
	OtherResult *checkResult                 `json:"otherResult,omitempty" yaml:"otherResult,omitempty"`
```

cmd/check.go
```diff
@@ -0,0 +1,176 @@
...
+	err = yaml.Unmarshal(reportBytes, &report)
+	if err != nil {
+		return result, fmt.Errorf("unmarshalling report file contents: %w", err)
+	}
+
+	// TODO: use different profiles based on command line argument
+	var p *profiles.Profile
+	switch profile {
+	case partnerProfile:
+		p = profiles.Get()
```
isutton, line 73: `p`'s assignment can happen outside the `switch` statement.

zonggen:
Thanks for reviewing @isutton ! After discussing with @mmulholla, he decides to merge this part into his new `report` command to integrate all report related parsing.

PR #161, opened 2021-06-28
add multi-profile support

mmulholla:
Added support for multiple profile. Current 3 profiles in new /config directory. One for partners, one for redhat and one for community. The content of these profiles is subject to change pending discussions. These file are include u=in the docker image under /app/config directory.

The default profile is partners version 1.1.

To specify which profile to use the --set flag:

* ```--set profile.vendorType=partner```     
  * valid values based on current profiles: partner, community, redhat, default
  * default is same as partner. 
  * If value specified is not recognized default will be assumed
* ```--set profile.version=v1.0```
  * Current only valid version is v1.0
  * If value specified is invalid v1.0 will be assumed. 

The report includes new profile attributes in the tool metadata:
```
    tool:
        verifier-version: 1.1.0
        profile:
            VendorType: community
            version: v1.0
        chart-uri: https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz?raw=true
        digest: sha256:3cda4165a30e77372332c6da2fb81f33ddfd83e1b9294090063c90f712b49ca1
        lastCertifiedTimestamp: "2021-06-25T20:47:08.954611+00:00"
        certifiedOpenShiftVersions: N/A
```

As a result of these changes I have bumped the verifier version from 1.0.0 to 1.1.0

Created #162 for README updates.

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -34,7 +34,7 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
-var Version = "1.0.1"
+var Version = "1.1.0"
 
 func init() {
 	allChecks = chartverifier.DefaultRegistry().AllChecks()
@@ -77,8 +77,8 @@ func filterChecks(set profiles.FilteredRegistry, subset []string, setEnabled boo
 	return selected, nil
 }
 
-func buildChecks(all checks.DefaultRegistry, enabled, disabled []string) (chartverifier.FilteredRegistry, error) {
-	profileChecks := profiles.Get().FilterChecks(all)
+func buildChecks(all checks.DefaultRegistry, config *viper.Viper, enabled, disabled []string) (chartverifier.FilteredRegistry, error) {
+	profileChecks := profiles.New(Version, config).FilterChecks(all)
 	switch {
 	case len(enabled) > 0 && len(disabled) > 0:
 		return nil, errors.New("--enable and --disable can't be used at the same time")
@@ -113,10 +113,6 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		Short: "Verifies a Helm chart by checking some of its characteristics",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			checks, err := buildChecks(allChecks, enabledChecksFlag, disabledChecksFlag)
-			if err != nil {
-				return err
-			}
 
 			// vals is a resulting map considering all the options the user has given.
 			vals, err := opts.MergeValues(getter.All(settings))
@@ -124,12 +120,18 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
 				return err
 			}
 
-			verifier, err := chartverifier.
-				NewVerifierBuilder().
+			verifierBuilder := chartverifier.NewVerifierBuilder().
 				SetValues(vals).
-				SetChecks(checks).
 				SetConfig(config).
-				SetOverrides(verifyOpts.Values).
+				SetOverrides(verifyOpts.Values)
+
+			checks, err := buildChecks(allChecks, verifierBuilder.GetConfig(), enabledChecksFlag, disabledChecksFlag)
+			if err != nil {
+				return err
+			}
+
+			verifier, err := verifierBuilder.
+				SetChecks(checks).
 				SetToolVersion(Version).
 				SetOpenShiftVersion(openshiftVersionFlag).
 				Build()
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -78,7 +78,7 @@ func filterChecks(set profiles.FilteredRegistry, subset []string, setEnabled boo
 }
 
 func buildChecks(all checks.DefaultRegistry, config *viper.Viper, enabled, disabled []string) (chartverifier.FilteredRegistry, error) {
-	profileChecks := profiles.New(Version, config).FilterChecks(all)
+	profileChecks := profiles.New(config).FilterChecks(all)
 	switch {
 	case len(enabled) > 0 && len(disabled) > 0:
 		return nil, errors.New("--enable and --disable can't be used at the same time")
--- a/pkg/chartverifier/profiles/profile.go
+++ b/pkg/chartverifier/profiles/profile.go
@@ -1,14 +1,10 @@
 package profiles
 
 import (
-	"errors"
-	"fmt"
-	"github.com/Masterminds/semver"
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
 	"github.com/spf13/viper"
+	"golang.org/x/mod/semver"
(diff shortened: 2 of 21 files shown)


mmulholla:
Need to figure out why test are failing, they pass locally

mmulholla:
Fixed the build issue. The build directory structure includes two ```chart-verifier`` directories which messed up the logic for finding the profiles. Fixed now.

pkg/chartverifier/profiles/profile.go
```diff
@@ -2,27 +2,44 @@ package profiles
 
 import (
 	"errors"
+	"fmt"
+	"github.com/Masterminds/semver"
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
+	"github.com/spf13/viper"
 	"gopkg.in/yaml.v3"
+	"io"
+	//"io"
```
zonggen, line 11: Redundant line

pkg/chartverifier/profiles/profile.go
```diff
@@ -98,3 +157,37 @@ func (profile *Profile) FilterChecks(registry checks.DefaultRegistry) FilteredRe
 	return filteredChecks
 
 }
+
+func readProfile(fileName string) (*Profile, error) {
+
+	// Open the json file which defines the tests to run
```
zonggen, line 163: s/json/yaml

pkg/chartverifier/reportBuilder.go
```diff
@@ -64,8 +64,9 @@ func (r *reportBuilder) SetToolVersion(version string) ReportBuilder {
 	return r
 }
 
-func (r *reportBuilder) SetProfile(profileName string) ReportBuilder {
-	r.Report.Metadata.ToolMetadata.Profile = profileName
+func (r *reportBuilder) SetProfile(vendorType profiles.VendorType, version string) ReportBuilder {
+	r.Report.Metadata.ToolMetadata.Profile.VendorType = string(vendorType)
```
zonggen, line 68: I think `string()` is redundant here since `profiles.VendorType` is already `string`

pkg/chartverifier/profiles/profile_test.go
```diff
@@ -4,80 +4,131 @@ import (
...
 func TestProfile(t *testing.T) {
 
 	testProfile := getDefaultProfile("test")
-	testProfile.Name = "profile-1.0.0"
+	testProfile.Name = "profile-partner-1.1"
+	config := viper.New()
 
 	t.Run("Profile read from disk should match test profile", func(t *testing.T) {
-		diskProfile := Get()
+		diskProfile := New("1.1.0", config)
```
zonggen, line 32: This should be "1.1" right? I'm kind of confused by the naming of `configVersion`, `checkVersion`, `profileVersion` and `cmd.Version`..

zonggen:
Sucessfully built + ran verifier with `--set` options with `podman run --rm quay.io/redhat-certification/chart-verifier:960a2b6 verify --set verifier.vendorType=partner --set verifier.version=1.1 https://github.com/zonggen/charts/releases/download/hashicorp-vault-0.12.2/hashicorp-vault-0.12.2.tgz`

pkg/chartverifier/reportBuilder.go
mmulholla: If i remove string it complains: Cannot use 'vendorType' (type profiles.VendorType) as type string.

pkg/chartverifier/profiles/profile_test.go
mmulholla: cmd.Version is the chart-verifier version - it is major.minor.patch, if we publish a new chart verifier with bug fixes we update the patch level, if we change behavior or add significant function we bump the minor version.

checkVersion is the version of a check, I didn't want to get into patch versions because we would get into check version hell - consider 1.0 as 1.0.any. if we update a check in a way which it can change an outcome for the same input we would bump it to 1.1 etc. Also conisder in terms of patch levels, we would always want the latest patch version so being able to specify older ones is redundant.

profileVersion - my original  intent was the profile version would mirror the chart-verifier version and we would issue a new one with each minor version update of the chart verifier. This breaks down and they need to be independent but the logic applied to checks, also applies here. 

configVersion is a profile version set through command line config when invoking the verifier. Maybe I could have picked a better name, but maybe not.

func New(version string, config *viper.Viper) *Profile - this function doesn't care whether it gets two digits or three, it just ignores the third digit. I did it this way because when the verify command calls New and it has a 3 digit version and it didn't want it to include logic to strip the 3rd digit. As a result the test sends 3 digits too. 

Hope this helps.

pkg/chartverifier/profiles/profile.go
mmulholla: removed.

pkg/chartverifier/profiles/profile.go
mmulholla: fixed

mmulholla:
Just added a commit with some extensive changes to how profiles are handled.

1. Changed profile versions to 1.0 as they no longer need to align with chart verifier version.
2. During init - walk the config directory and load all profiles found, save each profile in a map: ```var profileMap map[VendorType][]*Profile```
   - there can be more than one profile per vendor type.
   - if the version is not valid v1.0 is assumed
   - if the vendor type is not found "vendorTypeNotSpecified" is used
   - the default profile is added the map with a vendorType of "default" 
      - unless a profile with vendorType "default" was already loaded
      - vendor type in the profile is ```partner```

When choosing a profile base on --set profile.vendorType and --set profile.Version
- if the vendor type is found but the version is not, the vendor profile with the latest version is used.
- if vendor type is specified but not the version, the  vendor profile with the latest version in used.
- if the vendor type is not specified "default" is assumed.
- if just version is specified. vendor type "default" is assumed and the version is matched to any default profiles, if the version is not found the latest version of the default profile will be used.
- if neither are set the latest default profile is used.

The is designed to make it easy for profiles to be added and used by red hat or any other user, including setting an alternative default.

zonggen:
Minor: might be better to rename `reportBuilder.go` to `reportbuilder.go` as `verifierbuilder.go` is also all lower cased.

cmd/verify.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
-var Version = "1.0.1"
+var Version = "1.1.0"
```
dperaza4dustbit, line 37: Should this be a build property instead of being hardcoded here. For now OK just want to know if there is a plan to be more dynamic with this version.

dperaza4dustbit (approved):
Looks good to me, just one question on the engine verison

cmd/verify.go
mmulholla: This was made a var so it could be changed as part of a build.

go build -ldflags="-X 'github.com/redhat-certification/chart-verifier/cmd.Version=1.1.0'"

But no support yet for it to be dynamic.

PR #163, opened 2021-06-28
Sha values

mmulholla:
Added package sha to the report for tgz charts. The code for calculating the sha value was basically copied from helm code:

- https://github.com/helm/helm/blob/eb99434597d230a2106b22f15ac1e250ca5592f6/pkg/provenance/sign.go#L391
- https://github.com/helm/helm/blob/eb99434597d230a2106b22f15ac1e250ca5592f6/pkg/provenance/sign.go#L403

The package sha is added in the report:
```
        digest: sha256:0f84c30a7f18cf9adcfc5481b59755a71b555c8f420b41dc49360c4ba473892e
        digests:
            chart: sha256:0f84c30a7f18cf9adcfc5481b59755a71b555c8f420b41dc49360c4ba473892e
            package: 97e274069d9d3d028903610a3f9fca892b2620f0a334de6215ec5f962328586f
```

```digest:``` is left for backwards compatability and can be removed in the future.

If the report was run against chart src the digests.package attribute will be omitted. 

Tests included check that the calculated package sha matches the value return from ```sha256sum``` for local charts and ```curl -L <url> | sha256sum" for remote charts.

Also I verified the sha's caclulated for the currently published charts match the values currently in https://github.com/openshift-helm-charts/charts/blob/gh-pages/index.yaml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,4 +31,6 @@ COPY --from=build /tmp/src/out/chart-verifier /app/chart-verifier
 
 COPY --from=build /usr/local/bin/* /usr/local/bin/
 
+COPY ./config /app/config
+
 ENTRYPOINT ["/app/chart-verifier"]
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -34,7 +34,7 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier"
 )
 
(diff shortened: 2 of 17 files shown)


mmulholla:
replace by https://github.com/redhat-certification/chart-verifier/pull/164 after cherry-picking.

PR #164, opened 2021-06-28
Calculate package sha values

mmulholla:
Added package sha to the report for tgz charts. The code for calculating the sha value was basically copied from helm code:

- https://github.com/helm/helm/blob/eb99434597d230a2106b22f15ac1e250ca5592f6/pkg/provenance/sign.go#L391
- https://github.com/helm/helm/blob/eb99434597d230a2106b22f15ac1e250ca5592f6/pkg/provenance/sign.go#L403

The package sha is added in the report:
```
        digest: sha256:0f84c30a7f18cf9adcfc5481b59755a71b555c8f420b41dc49360c4ba473892e
        digests:
            chart: sha256:0f84c30a7f18cf9adcfc5481b59755a71b555c8f420b41dc49360c4ba473892e
            package: 97e274069d9d3d028903610a3f9fca892b2620f0a334de6215ec5f962328586f
```

```digest:``` is left for backwards compatability and can be removed in the future.

If the report was run against chart src the digests.package attribute will be omitted. 

Tests included check that the calculated package sha matches the value return from ```sha256sum``` for local charts and ```curl -L <url> | sha256sum" for remote charts.

Also I verified the sha's caclulated for the currently published charts match the values currently in https://github.com/openshift-helm-charts/charts/blob/gh-pages/index.yaml

### DIFF
--- a/pkg/chartverifier/report.go
+++ b/pkg/chartverifier/report.go
@@ -49,7 +49,7 @@ type ToolMetadata struct {
 	Version                    string  `json:"verifier-version" yaml:"verifier-version"`
 	Profile                    string  `json:"profileName" yaml:"profileName"`
 	ChartUri                   string  `json:"chart-uri" yaml:"chart-uri"`
-	Digest                     string  `json:"digest,omitempty" yaml:"digest"`
+	Digest                     string  `json:"digest" yaml:"digest"`
 	Digests                    Digests `json:"digests" yaml:"digests"`
 	LastCertifiedTimestamp     string  `json:"lastCertifiedTimestamp,omitempty" yaml:"lastCertifiedTimestamp,omitempty"`
 	CertifiedOpenShiftVersions string  `json:"certifiedOpenShiftVersions,omitempty" yaml:"certifiedOpenShiftVersions,omitempty"`
--- a/pkg/chartverifier/report.go
+++ b/pkg/chartverifier/report.go
@@ -46,12 +46,18 @@ type ReportMetadata struct {
 }
 
 type ToolMetadata struct {
-	Version                    string `json:"verifier-version" yaml:"verifier-version"`
-	Profile                    string `json:"profileName" yaml:"profileName"`
-	ChartUri                   string `json:"chart-uri" yaml:"chart-uri"`
-	Digest                     string `json:"digest,omitempty" yaml:"digest,omitempty"`
-	LastCertifiedTimestamp     string `json:"lastCertifiedTimestamp,omitempty" yaml:"lastCertifiedTimestamp,omitempty"`
-	CertifiedOpenShiftVersions string `json:"certifiedOpenShiftVersions,omitempty" yaml:"certifiedOpenShiftVersions,omitempty"`
+	Version                    string  `json:"verifier-version" yaml:"verifier-version"`
+	Profile                    string  `json:"profileName" yaml:"profileName"`
+	ChartUri                   string  `json:"chart-uri" yaml:"chart-uri"`
+	Digest                     string  `json:"digest,omitempty" yaml:"digest"`
+	Digests                    Digests `json:"digests" yaml:"digests"`
+	LastCertifiedTimestamp     string  `json:"lastCertifiedTimestamp,omitempty" yaml:"lastCertifiedTimestamp,omitempty"`
+	CertifiedOpenShiftVersions string  `json:"certifiedOpenShiftVersions,omitempty" yaml:"certifiedOpenShiftVersions,omitempty"`
+}
+
+type Digests struct {
+	Chart   string `json:"chart" yaml:"chart"`
+	Package string `json:"package,omitempty" yaml:"package,omitempty"`
 }
 
 type CheckReport struct {
--- a/pkg/chartverifier/reportBuilder.go
+++ b/pkg/chartverifier/reportBuilder.go
@@ -17,10 +17,18 @@
 package chartverifier
 
 import (
+	"crypto"
 	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"github.com/pkg/errors"
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/profiles"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/checks"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pkg/chartverifier/reportbuilder_test.go [hunk 1]
```diff
@@ -43,3 +48,53 @@ func TestSha(t *testing.T) {
...
+		"checks/chart-0.1.0-v3.without-readme.tgz",
+		"checks/chart-0.1.0-v3.with-crd.tgz"}
+
+	for _, chart := range charts {
+		cmd := exec.Command("sha256sum", chart)
+		var out bytes.Buffer
+		cmd.Stdout = &out
+		err := cmd.Run()
+		if err != nil {
+			fmt.Println("error running command:")
```
zonggen, line 66: We should probably report the err in the case

pkg/chartverifier/report.go
```diff
@@ -46,12 +46,18 @@ type ReportMetadata struct {
...
-	Version                    string `json:"verifier-version" yaml:"verifier-version"`
-	Profile                    string `json:"profileName" yaml:"profileName"`
-	ChartUri                   string `json:"chart-uri" yaml:"chart-uri"`
-	Digest                     string `json:"digest,omitempty" yaml:"digest,omitempty"`
-	LastCertifiedTimestamp     string `json:"lastCertifiedTimestamp,omitempty" yaml:"lastCertifiedTimestamp,omitempty"`
-	CertifiedOpenShiftVersions string `json:"certifiedOpenShiftVersions,omitempty" yaml:"certifiedOpenShiftVersions,omitempty"`
+	Version                    string  `json:"verifier-version" yaml:"verifier-version"`
+	Profile                    string  `json:"profileName" yaml:"profileName"`
+	ChartUri                   string  `json:"chart-uri" yaml:"chart-uri"`
+	Digest                     string  `json:"digest,omitempty" yaml:"digest"`
```
zonggen, line 52: Is there a reason we only `omitempty` for json but not for yaml?

pkg/chartverifier/reportbuilder_test.go
mmulholla: added some assert.NoError checks in the tests.

pkg/chartverifier/report.go
mmulholla: Good spot, removed it from both now.

pkg/chartverifier/reportBuilder.go
```diff
@@ -150,3 +162,43 @@ func GenerateSha(rawFiles []*helmchart.File) string {
...
+	if err != nil {
+		return ""
+	}
+	var chartReader io.Reader
+	switch url.Scheme {
+	case "http", "https":
+		var chartGetResponse *http.Response
+		chartGetResponse, err = http.Get(url.String())
+		if err == nil {
+			chartReader = chartGetResponse.Body
```
dperaza4dustbit, line 178: Lets say I expect url to be valid and this still fails. Is there a way to see this error? Also do you happen to know if http.Get() support redirects? Have you tested with IBM raw paths?

pkg/chartverifier/reportBuilder.go
```diff
@@ -102,6 +111,9 @@ func (r *reportBuilder) Build() (*Report, error) {
 			}
 		}
 	}
+
+	r.Report.Metadata.ToolMetadata.Digests.Package = GetPackageDigest(r.Report.Metadata.ToolMetadata.ChartUri)
```
dperaza4dustbit, line 115: Is this going to show digest: "" then if there was an issue? Should we check first if GetPackageDigest return non empty?

pkg/chartverifier/reportBuilder.go
```diff
@@ -150,3 +162,43 @@ func GenerateSha(rawFiles []*helmchart.File) string {
...
+		chartGetResponse, err = http.Get(url.String())
+		if err == nil {
+			chartReader = chartGetResponse.Body
+		}
+	case "file", "":
+		if strings.HasSuffix(url.Path, ".tgz") {
+			chartReader, _ = os.Open(url.Path)
+		}
+	default:
+		err = errors.Errorf("scheme %q not supported", url.Scheme)
```
dperaza4dustbit, line 185: Can you think of an example for this non supported scheme?

pkg/chartverifier/reportbuilder_test.go
```diff
@@ -43,3 +48,53 @@ func TestSha(t *testing.T) {
...
+
+	charts := []string{"https://github.com/redhat-certification/chart-verifier/blob/main/pkg/chartverifier/checks/chart-0.1.0-v3.valid.tgz?raw=true",
+		"https://github.com/openshift-helm-charts/charts/releases/download/hashicorp-vault-0.13.0/hashicorp-vault-0.13.0.tgz",
+		"https://github.com/openshift-helm-charts/charts/releases/download/hashicorp-vault-0.12.0/hashicorp-vault-0.12.0.tgz",
+		"https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.1.2.tgz?raw=true"}
+
+	for _, chart := range charts {
+
+		var out bytes.Buffer
+		cmd1 := exec.Command("curl", "-L", chart)
```
dperaza4dustbit, line 82: Why depend on curl for the test? We can hardcode the digest right? simple map would do I think.

dperaza4dustbit (commented):
Since we are exposing a new thing in the report should we mention in our docs too

pkg/chartverifier/reportBuilder.go
mmulholla: This is one weakness of the cobra command structure we only output a message if the command fails and no report results. One suggestion I have is to add a messages attribute to the tool metadata where we include such errors.

pkg/chartverifier/reportBuilder.go
mmulholla: No, tested that "" counts as empty for omitempty.

pkg/chartverifier/reportBuilder.go
mmulholla: I took the lead here from https://github.com/redhat-certification/chart-verifier/blob/a857b7d4e6a393f8ac09339eaa9e31fc773cd6e8/pkg/chartverifier/checks/helm.go#L161

I guess in theory the user could enter anything although we probably would then fail way before we go to this code. Also difficult to write code which doesn't cover unexpected values.

pkg/chartverifier/reportbuilder_test.go
mmulholla: we could but I baulked at that because I didn't want the test to start failing if the target chart changed for any reason, which is quite likely for our own test chart.

pkg/chartverifier/reportBuilder.go
dperaza4dustbit: OK perfect thanks

pkg/chartverifier/reportBuilder.go
dperaza4dustbit: Just curious if you though of scenarios like that, but yeah is better to be save and check for unexpected values... defensive is good here.

pkg/chartverifier/reportbuilder_test.go
dperaza4dustbit: If test chart change we change the she right, and the chart will be a diff version. Are you saying that we will repackage the same version? Why at the end is just a binary we want to test. We might change the location to test in diff circumstance but binary can be constant, in fact id does not have to be a full blown chart is could be a little one we push.

dperaza4dustbit: I rather do not depend on underlying OS, that seems to be the best practice with go trying to build all in same executable. What if we run on a linux where curl is not available we can still unit test.

mmulholla: Updated the test to use hard coded sha values

pkg/chartverifier/reportBuilder.go
dperaza4dustbit: Like a verbose run? We can do that in

dperaza4dustbit (approved):

PR #169, opened 2021-07-07
Report to include correct profile

mmulholla:
Was overusing a variable ```profile``` used to keep track of the profile in use. Have renamed in to ```profileInUse``` to prevent overuse. Also extended a test to cover the case where it was breaking.

### DIFF
--- a/pkg/chartverifier/profiles/profile.go
+++ b/pkg/chartverifier/profiles/profile.go
@@ -91,7 +91,7 @@ func New(config *viper.Viper) *Profile {
 
 	profileInUse = getDefaultProfile(fmt.Sprintf("profile %s not found", profileVendorType))
 
-	if vendorProfiles,ok := profileMap[profileVendorType]; ok {
+	if vendorProfiles, ok := profileMap[profileVendorType]; ok {
 		profileInUse = vendorProfiles[0]
 		if len(vendorProfiles) > 1 {
 			for _, vendorProfile := range vendorProfiles {
--- a/pkg/chartverifier/profiles/profile.go
+++ b/pkg/chartverifier/profiles/profile.go
@@ -89,12 +89,11 @@ func New(config *viper.Viper) *Profile {
 		}
 	}
 
-	vendorProfiles := profileMap[profileVendorType]
-	if len(vendorProfiles) > 0 {
-		profileInUse = vendorProfiles[0]
+	profileInUse = getDefaultProfile(fmt.Sprintf("profile %s not found", profileVendorType))
 
+	if vendorProfiles,ok := profileMap[profileVendorType]; ok {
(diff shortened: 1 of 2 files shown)


pkg/chartverifier/profiles/profile.go
```diff
@@ -88,30 +89,30 @@ func New(config *viper.Viper) *Profile {
...
-			}
-			if semver.Compare(semver.MajorMinor(vendorProfile.Version), semver.MajorMinor(defaultProfile.Version)) > 1 {
-				defaultProfile = vendorProfile
 			}
 		}
 	}
-
-	if profile == nil {
-		profile = defaultProfile
+	if profileInUse == nil {
```
dperaza4dustbit, line 111: I would initialize profileInUse to this first then do the logic of the mapping to override or not.

pkg/chartverifier/profiles/profile.go
```diff
@@ -88,30 +89,30 @@ func New(config *viper.Viper) *Profile {
...
-		profile = nil
-		for _, vendorProfile := range vendorProfiles {
-			if len(profileVersion) > 0 {
-				if semver.Compare(semver.MajorMinor(vendorProfile.Version), semver.MajorMinor(profileVersion)) == 0 {
-					profile = vendorProfile
-					break
+	if len(vendorProfiles) > 0 {
+		profileInUse = vendorProfiles[0]
+
+		if len(vendorProfiles) > 1 {
```
dperaza4dustbit, line 96: Instead of nesting like this can we we just check above if len(vendorProfiles) == 1 we set to vendorProfiles[0]. Then this will be just an els if but this time checking for  len(vendorProfiles) > 1. No need to set to nil cause you will default to default profile above this like

dperaza4dustbit (commented):
A few suggestions to flatten the nested if block

pkg/chartverifier/profiles/profile.go
mmulholla: checking for len being greater than 0 protects against profile.vendorype being set to an invalid value which means there are no vendor profiles in the map.

mmulholla: Changed it to use:
``` 
	if vendorProfiles,ok := profileMap[profileVendorType]; ok {
             ......
        } 
```

to be clearer

pkg/chartverifier/profiles/profile.go
mmulholla: Done

pkg/chartverifier/profiles/profile.go
dperaza4dustbit: OK looks good

dperaza4dustbit (approved):

PR #174, opened 2021-07-16
add chart-uri and chart to report command metadata output

mmulholla:
added chart-uri and chart metdata to the outout of report metadata, for example: 
```
metadata:
    vendorType: redhat
    profileVersion: v1.0
    chart-uri: ../../test/ibm-oms-pro-prod-6.0.1.tgz
    chart:
        name: ibm-oms-pro-prod
        home: http://ibm.biz/oms-home
        sources: []
        version: 6.0.1
       etc...
```

This addresses 2 problems.
1. workflow uses chart-uri and chart.name and chart.version from the verifier report and as a result hard codes to the verifier report format. Using the report command makes the workflow independant of verifier report format.
2. Because the workflow works based on json format and by getting the chart data from the report command in json format, empty fields are not included  which automagically fixes the importvalues problem: https://issues.redhat.com/browse/HELM-210

### DIFF
--- a/cmd/report_test.go
+++ b/cmd/report_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/redhat-certification/chart-verifier/pkg/chartverifier/report"
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/require"
+	helmchart "helm.sh/helm/v3/pkg/chart"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


dperaza4dustbit (approved):

zonggen (approved):
> empty fields are not included which automagically fixes the importvalues problem

:+1:

PR #176, opened 2021-08-02
add new version support: prep for release automation

mmulholla:
Add support for reading the chart verifier version from a json file in preparation for release automation.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,1 +1,68 @@
 package cmd
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+var Version = "0.0.0"
+
+type Release struct {
+	Version string `json:"version"`
+}
+
+func init() {
+
+	var configDir string
+	if isRunningInDockerContainer() {
+		configDir = filepath.Join("app", "releases")
+	} else {
+		_, fn, _, ok := runtime.Caller(0)
+		if !ok {
+			return
+		}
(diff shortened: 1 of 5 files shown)


cmd/version.go
```diff
@@ -1 +1,68 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+var Version = "0.0.0"
```
zonggen, line 12: Minor: Is there a reason this is exported?

cmd/version_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/mod/semver"
+	"testing"
+)
+
+func TestVersion(t *testing.T) {
+
+	t.Run("Check Version is set.", func(t *testing.T) {
+		fmt.Println(fmt.Sprintf("Version is %s", Version))
```
zonggen, line 13: Minor: linter suggests: `fmt.Printf("Version is %s\n", version)`

cmd/version.go
mmulholla: Probably not while it is only used in verify.go although it is the source of truth for the version so like it being available.

cmd/version_test.go
mmulholla: Not a fan of "\n" in the text, prefer Println to take care of that.

dperaza4dustbit (approved):

zonggen (approved):

PR #188, opened 2021-08-27
add latest tag to new image in quay

mmulholla:
Added a worklfow step to add the latest tag a newly created chart verifier release image in quay. 

The image will not have been created when the step starts so the step checks up to 60 times, every 15 seconds for the image to appear. This is a total of 15 minutes. If this fails the image will have to be tagged manually, instructions in the README. The wait time may need adjusting based in it success rate.

Also note to add the tag an OAUTH token is required and is set as a github repository secret "QUAY_AUTH_TOKEN"

### DIFF
--- a/scripts/src/quay/linkimages.py
+++ b/scripts/src/quay/linkimages.py
@@ -1,0 +1,133 @@
+"""
+Used by a github action to link a newly published chart verifier image to a specified tag in quay
+Can also be called from the command line.
+
+parameters:
+    ---link-tag : tag to link the image to, default is "test"
+    --verifier-version : the version of the new chart verifier release
+
+notes:
+    It is anticipated that the github action will invoke this function just before the new image is available
+    in quay. As a result the function will look for the image every 15 seconds for up to 15 minutes.
+
+    To invoke from the command line, invoke from the root directory of the repository,for example:
+          python3 scripts/src/quay/linkimages.py --verifier-version=1.2.0.
+
+    If the verifier version image is already linked to the specified tag no action is taken.
+
+    The default of link-tag is set to "test" to avoid accidental updates to latest.
+
+    The auth token to enable the script to link tags in quay must be set in an environment variable QUAY_AUTH_TOKEN.
+    For a github action this must be set as a repository secret.
+    For information on auth token required see:
+    https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_api_guide/using_the_red_hat_quay_api
+
+results:
+    A message indicating the outcome.
+    exit code 1 if version image was not found.
+
+"""
+
+import requests
+import json
+import sys
+import argparse
+import os
+from retry import retry
+
+sys.path.append('./scripts/src/')
+from release import releasechecker
+
+# Quay API docs: https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_api_guide/index
+# Quay Swagger API: https://docs.quay.io/api/swagger/
+#
+
+defaultLinkTag= "test"
+tagUrl = 'https://quay.io/api/v1/repository/redhat-certification/chart-verifier/tag/'
+
+# try every 15 seconds for 15 minutes
+@retry(Exception,tries=60, delay=15)
+def getImageId(tagValue,doRetry):
+
+    print(f"[INFO] look for tag : {tagValue}, retry : {doRetry}")
+    tagUrl = 'https://quay.io/api/v1/repository/redhat-certification/chart-verifier/tag/'
+
+    get_params = {'onlyActiveTags' : 'true','specificTag' : tagValue }
+
+    response = requests.get(tagUrl,params=get_params)
+
+    imageId = ""
+    if response.status_code > 201:
+        print(f"[Error] Error getting tags from quay : status_code={response.status_code}")
+    else:
+        tags = json.loads(response.text)
+        print("[INFO] loaded the tags")
+        for tag in tags["tags"]:
+            if tag['name'] == tagValue:
+                imageId = tag['image_id']
+                print(f"[INFO] Found tag {tagValue}. image_id : {imageId}")
+                break
+            else:
+                print(f"[INFO] ignore tag {tag['name']}")
+
+        if not imageId and doRetry:
+            print(f"[INFO] {tagValue} not found. Retry!")
+            raise Exception(f"Image {tagValue} not found")
+
+    return imageId
+
+def linkImage(linkImage,linkTag):
+
+    print(f"[INFO] Update {linkTag} to point to {linkImage}")
+    auth_token = os.environ.get('QUAY_AUTH_TOKEN')
+    if auth_token is None:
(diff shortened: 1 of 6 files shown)


scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+                print(f"[INFO] Found tag {tagValue}. image_id : {imageId}")
+                break
+            else:
+                print(f"[INFO] ignore tag {tag['name']}")
+
+        if not imageId and doRetry:
+            print(f"[INFO] {tagValue} not found. Retry!")
+            raise Exception(f"Image {tagValue} not found")
+
+    return imageId
```
dperaza4dustbit, line 77: Do you want to return empty on error? If you do then why printing error and not warning?

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+
+def linkImage(linkImage,linkTag):
+
+    print(f"[INFO] Update {linkTag} to point to {linkImage}")
+    auth_token = os.environ.get('QUAY_AUTH_TOKEN')
```
dperaza4dustbit, line 82: Who does that token belong to? Your Identity?

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+    if auth_token is None:
```
dperaza4dustbit, line 83: cleaner to say just: if not auth_token

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+        print("[ERROR] repository secret QUAY_AUTH_TOKEN not set")
+        return False
+
+    quay_token = f"Bearer {auth_token}"
+    put_header = {'content-type': 'application/json','Authorization': quay_token}
+
+    puturl = tagUrl + linkTag
```
dperaza4dustbit, line 90: could also use the f"{tagUrl}{linkTag" here

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+    putData = {'image': linkImage}
+    putOut = requests.put(puturl,data=json.dumps(putData), headers=put_header)
+    print(f"[INFO] Update link response code : {putOut.status_code}")
+    print(f"[INFO] Update link response : {putOut.text}")
+
+    return putOut.status_code == 200 or putOut.status_code == 201
```
dperaza4dustbit, line 96: could use: return putOut.status_code in [200, 201]

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+
+# Quay API docs: https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_api_guide/index
+# Quay Swagger API: https://docs.quay.io/api/swagger/
+#
+
+defaultLinkTag= "test"
+tagUrl = 'https://quay.io/api/v1/repository/redhat-certification/chart-verifier/tag/'
+
+# try every 15 seconds for 15 minutes
+@retry(Exception,tries=60, delay=15)
```
dperaza4dustbit, line 49: This is very cool decorator

scripts/src/release/releasechecker.py
```diff
@@ -67,6 +67,11 @@ def make_release_body(version, image_name, release_info):
     print(f"[INFO] Release body: {body}")
     print(f"::set-output name=PR_release_body::{body}")
 
+def get_version_info():
+    file = open(VERSION_FILE,)
```
dperaza4dustbit, line 71: comma at the end? I have seen that but prefer you specify the mode explicitly "r" in this case.  one way the I like loading json is this:
data = {}
with open('data.json') as json_file:
    data = json.load(json_file)
return data
This with will handle the opening and closing of the handle.

dperaza4dustbit (commented):
Logic is solid, just a few python specific suggestions. Try using pep8 tool, will tell you all things to follow conventions.

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+def getImageId(tagValue,doRetry):
```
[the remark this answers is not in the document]
mmulholla: Update all methods and variables to match the conventions.

scripts/src/quay/linkimages.py
```diff
@@ -0,0 +1,133 @@
...
+    print(f"[INFO] look for tag : {tagValue}, retry : {doRetry}")
+    tagUrl = 'https://quay.io/api/v1/repository/redhat-certification/chart-verifier/tag/'
+
+    get_params = {'onlyActiveTags' : 'true','specificTag' : tagValue }
+
+    response = requests.get(tagUrl,params=get_params)
+
+    imageId = ""
+    if response.status_code > 201:
+        print(f"[Error] Error getting tags from quay : status_code={response.status_code}")
```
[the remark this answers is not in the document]
mmulholla: If I throw an exception the retry mechanism will kick in. I did not want to force retries for 15 minutes if we got a 404 or 500 - which would probably be a sign of other issues - so just fall back on the manual process.

scripts/src/quay/linkimages.py
mmulholla: The method is used to look for the latest tag as well as the new n.n.n tag. When looking up latest it would not be be error if it did not exist yet. I did add some logic to the caller to error if an empty string is returned for the n.n.n tag although in theory that should always result in an exception.

scripts/src/quay/linkimages.py
mmulholla: Yes it does, but we should all be able to create and replace it. I do include some information about it both in the comments in linkimages.py and in the README. I think we could set up a Robot ID to get it (I think) but given anyone can create one I don't see a need.

scripts/src/quay/linkimages.py
mmulholla: Done

scripts/src/quay/linkimages.py
mmulholla: Done

scripts/src/quay/linkimages.py
mmulholla: Much nicer, thank you.

scripts/src/quay/linkimages.py
mmulholla: Agreed :-)

scripts/src/release/releasechecker.py
mmulholla: much neater. Used your logic.

scripts/src/quay/linkimages.py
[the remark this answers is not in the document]
mmulholla: also updated the condtional, much nicer

mmulholla:
@dperaza4dustbit - responses added to all comments

scripts/src/quay/linkimages.py
dperaza4dustbit: OK

dperaza4dustbit (approved):
lgtm

PR #193, opened 2021-09-08
Go 1.16; dep updates

sebrandon1:
- Go 1.16.7 is the latest available 1.16 [version](https://golang.org/dl/).
- Went through and did some `go get -u` updates to various dependencies.
- Updated `helm-chart-developer.md` to reflect new Go version and SHA.
- Per the [1.16 release notes](https://golang.org/doc/go1.16), ioutil library has some functions moved/deprecated.  Updated to reflect those changes.


sebrandon1:
Interesting, no matter how I change the `.github/workflows/Build.yaml` it still tries to checkout the `upstream/main` version of it which is still stuck to Go 1.15.8.  Any idea on why this is happening?

mmulholla:
The build.yml is run from the PR target to prevent malicious updates. It only runs certain steps using the pr branch.  You need to test on your fork by merging into main on your fork and then testing, and add a link to this PR to show the  results.

Can I also ask why you are making this change?

PR #194, opened 2021-09-16
Initial test features for acceptance test

mmulholla:
Initial attempt at defining first pass of new acceptance tests - golden path only.


tests/features/chart_source_local.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Chart local source verification
+    Partners or redhat or community can verify their charts by running the
+    chart verifier against a local chart source
```
zonggen, line 3:
```suggestion
    chart verifier against a local chart source.
```

Official doc added period at the end of feature description: https://github.com/pytest-dev/pytest-bdd#example

zonggen (commented):
From the documentation (https://github.com/pytest-dev/pytest-bdd#example),

- Given: I am a Red Hat partner from IBM
  - And: I would like to run chart verifier using the docker image
  - And: I would like to use the partner profile
- When: I run the "chart-verifier xxxx" comand
- Then: I should see a report which passes all mandatory tests

tests/functional/features/chart_source_local.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Chart local source verification
+    Partners or redhat or community can verify their charts by running the
+    chart verifier against a local chart source.
+
+    Scenario: The partner IBM succesfully verifies their chart source using the chart verifier docker image
+        Given: I am a Red Hat partner from IBM
```
zonggen, line 6: Sorry that I used colon in my example because of markdown. But `Given` `And` `When` and `Then` sections do not have colons: https://github.com/pytest-dev/pytest-bdd#example

PR #195, opened 2021-09-17
Update ibm images which have been updated

mmulholla:
Two ibm certified images we use in testing appear to have been removed and newer version added. 

Have updated tests to use valid images. 

Long run we may want to add some of our own images.

Build an test step is failing because it uses the chart-0.1.0-v3.valid.tgz from the main branch and it is part of this update.


dperaza4dustbit (approved):
lgtm

PR #202, opened 2021-10-06
fix test image issues

mmulholla:
Some ibm images were stale, so updated tests which use them.

Also updated chart-0.1.0-v3.valid.tgz:
- to use newer ibm images
- increase flexibility for the kube version specified in chart.yaml.

Note: there is also a test which check the sha for this chart, but uses the chart from main. As a result this test will fail intil the new this PR is merged.


mmulholla:
> @mmulholla How do you get the expected error? Ran `make test` and all tests passed.

The failure only happens in the PR build, make test just runs go tests and the failing test is not part of that. If you look in the PR build logs you will see the error.

zonggen (approved):
Ah saw the logic in the `buildandtest.py`, makes sense now.

LGTM

PR #209, opened 2021-10-20
enable chart verifier to recognize OCP 4.9

mmulholla:
Chart verifier now recognizes kube version 1.22 as OCP version 4.9


pkg/tool/oc_test.go
```diff
@@ -1,41 +1,58 @@
...
 
-var output48 = `
-serverVersion:
-  major: "1"
-  minor: "21"
-`
-
-func TestOcVersion48(t *testing.T) {
-	fp := fakeProcessExecutor{Output: output48}
+func checkOcVersion(kubeVersion string, expectedOCversion string, t *testing.T) {
```
zonggen, line 51:
```suggestion
func checkOcVersion(kubeVersion string, expectedOCVersion string, t *testing.T) {
```

pkg/tool/oc_test.go
```diff
@@ -1,41 +1,58 @@
...
+	fp := fakeProcessExecutor{Output: kubeVersion}
 	oc := NewOc(fp)
 	version, _ := oc.GetVersion()
-	if version != "4.8.0" {
-		t.Error("Version mismatch", version)
+	if version != expectedOCversion {
+		t.Error(fmt.Sprintf("Version mismatch expected: %s, got: %s", expectedOCversion, version))
```
zonggen, line 56:
```suggestion
		t.Errorf("Version mismatch expected: %s, got: %s", expectedOCversion, version)
```

pkg/tool/oc_test.go
mmulholla: Updated

pkg/tool/oc_test.go
mmulholla: Updated

PR #213, opened 2021-10-28
[DO NOT MERGE] Add new kubeVersion check and annotation

mmulholla:
DO NOT MERGE, merge needs to sync with: https://github.com/openshift-helm-charts/development/pull/59

Adds a new ```v1.1/has-kubeversion``` check to the 1.1 profiles.

It ensure sthe chart kubeVersion is set to a valid semantic version. From this the supported OpenShiftVersions are determined and included in an annotation in the report.  

Also added to profile 1.1 the rename  of annotation : certifiedOpenShiftVersions to testedOpenShiftVersion


pkg/chartverifier/checks/checks.go
```diff
@@ -337,3 +361,40 @@ func parseImageReference(image string) pyxis.ImageReference {
...
+
+func getOCPRange(kubeVersionRange string) string {
+
+	semverCompare := sprig.GenericFuncMap()["semverCompare"].(func(string, string) (bool, error))
+	minOCPVersion := ""
+	maxOCPVersion := ""
+	for kubeVersion, OCPVersion := range tool.GetKubeOpenShiftVersionMap() {
+		match, err := semverCompare(kubeVersionRange, kubeVersion)
+		if err != nil {
+			return fmt.Sprintf("%s : %s", KuberVersionProcessingError, err)
```
zonggen, line 373: Nit: can we use the common style for error handling in golang: https://golang.org/doc/effective_go#multiple-returns? i.e. return result as well as the error if any (`func getOCPRange(kubeVersionRange string) string error`).

pkg/chartverifier/checks/checks.go
```diff
@@ -337,3 +361,40 @@ func parseImageReference(image string) pyxis.ImageReference {
...
+			if minOCPVersion == "" || semver.Compare(testOCPVersion, fmt.Sprintf("v%s", minOCPVersion)) < 0 {
+				minOCPVersion = OCPVersion
+			}
+			if maxOCPVersion == "" || semver.Compare(testOCPVersion, fmt.Sprintf("v%s", maxOCPVersion)) > 0 {
+				maxOCPVersion = OCPVersion
+			}
+		}
+	}
+	if minOCPVersion != "" {
+		match, _ := semverCompare(kubeVersionRange, "1.999")
```
zonggen, line 386: Confused in this condition check, why are we comparing to 1.999 here?

pkg/chartverifier/report/testreports/v1.0/reportallsorts.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
         digests:
             chart: sha256:0c1c44def5c5de45212d90396062e18e0311b07789f477268fbf233c1783dbd0
         lastCertifiedTimestamp: "2021-07-02T08:09:56.881793-04:00"
-        certifiedOpenShiftVersions: N/A
+        certifiedOpenShiftVersions : N/A
```
zonggen, line 14:
```suggestion
        certifiedOpenShiftVersions: N/A
```

pkg/chartverifier/checks/checks.go
zonggen: Got it. Thanks for explaining! I think a comment as a note will be helpful.

pkg/chartverifier/checks/checks.go
mmulholla: Fixed

pkg/chartverifier/report/testreports/v1.0/reportallsorts.yaml
mmulholla: Fixed

pkg/chartverifier/checks/checks.go
mmulholla: Comment now added.

dperaza4dustbit (approved):
lgtm

zonggen (approved):
LGTM!

PR #215, opened 2021-11-04
remove redundant digest metadata

mmulholla:
Remove redundant digest from metadata - was left for compatibility reasons and no longer needed.

See: https://github.com/redhat-certification/chart-verifier/pull/163


zonggen (approved):
LGTM!

PR #218, opened 2021-11-17
add option for user to specify helm release for testing

mmulholla:
Adds ability for a chart-verifer user to specify the release to run chart testing on:

```--set chart-testing.release=use-this-1```

this method of setting options for specific tests is used throughout. ie: ```<check-name>.<option-name>``` 

There are requirements on the value used for a release and if you get it wrong install fails, for example:
```
 Error: INSTALLATION FAILED: Service "funny.bone-chart" is invalid: metadata.name: Invalid value: "funny.bone-chart": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

As a result I did not add any checking of a specified  release name when I read it.

### DIFF
--- a/docs/helm-chart-checks.md
+++ b/docs/helm-chart-checks.md
@@ -248,6 +248,7 @@ You can configure the chart-testing check by performing one of the following ste
         --set chart-testing.skipMissingValues=true                    \
         --set chart-testing.namespace=${NAMESPACE}                    \
         --set chart-testing.releaseLabel="app.kubernetes.io/instance" \
+        --set chart-testing.release=${RELEASE}
         some-chart.tgz
     ```
 * Option 2: Create a YAML file (config.yaml) similar to the following example:
@@ -258,12 +259,15 @@ You can configure the chart-testing check by performing one of the following ste
         skipMissingValues: true
         namespace: <NAMESPACE>
         releaseLabel: "app.kubernetes.io/instance"
+        release: <RELEASE>
     ```
 
     Specify the file using the `--set-values` command line option:
     ```text
     $ chart-verifier verify --enable chart-testing --set-values config.yaml some-chart.tgz
     ```
+
+    All settings are optional, if not set a default values will be used.
 
 ### Override values
 
--- a/docs/helm-chart-checks.md
+++ b/docs/helm-chart-checks.md
@@ -267,7 +267,7 @@ You can configure the chart-testing check by performing one of the following ste
     $ chart-verifier verify --enable chart-testing --set-values config.yaml some-chart.tgz
     ```
 
-    All settings are optional, if not set a default values will be used.
+    All settings are optional, if not set default values will be used.
 
 ### Override values
 
--- a/pkg/chartverifier/checks/charttesting.go
+++ b/pkg/chartverifier/checks/charttesting.go
@@ -16,6 +16,10 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const (
+	ReleaseConfigString string = "release"
+)
+
 // Versioner provides OpenShift version
 type Versioner func() (string, error)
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: pkg/chartverifier/checks/charttesting.go [hunk 1]
```diff
@@ -107,6 +111,11 @@ func ChartTesting(opts *CheckOptions) (Result, error) {
 		return NewResult(false, err.Error()), nil
 	}
 
+	configRelease := opts.ViperConfig.GetString(ReleaseConfigString)
+	if len(configRelease) >0 {
```
zonggen, line 115:
```suggestion
	if len(configRelease) > 0 {
```

`go fmt` check is failing.

docs/helm-chart-checks.md
```diff
@@ -258,13 +259,16 @@ You can configure the chart-testing check by performing one of the following ste
...
         releaseLabel: "app.kubernetes.io/instance"
+        release: <RELEASE>
     ```
 
     Specify the file using the `--set-values` command line option:
     ```text
     $ chart-verifier verify --enable chart-testing --set-values config.yaml some-chart.tgz
     ```
 
+    All settings are optional, if not set a default values will be used.
```
zonggen, line 270:
```suggestion
    All settings are optional, if not set default values will be used.
```

pkg/chartverifier/checks/charttesting.go
```diff
@@ -160,16 +169,24 @@ func generateInstallConfig(
...
 		cleanup = func() {
 			helm.DeleteRelease(namespace, release)
 		}
 	} else {
-		release, namespace = chrt.CreateInstallParams(cfg.BuildId)
+		if len(release) == 0 {
+			release, namespace = chrt.CreateInstallParams(cfg.BuildId)
+		} else {
+			_, namespace = chrt.CreateInstallParams(cfg.BuildId)
+		}
```
zonggen, line 189: Seems the if blocks are duplicated, can you put the `len(release)` check at the begining of this function instead of repeating in each condition?

mmulholla: Blocks are subtly different in what they get back from chrt.CreateInstallParams(cfg.BuildId). There are 4 things to cover:
1. Namespace and Release both set
2. Namespace set and Release not set
3. Namespace not set and Release set
4. Neither Namespace nor Release set

chose nested ifs insteand of a if, else if, else if, else

docs/helm-chart-checks.md
mmulholla: thx, fixed

pkg/chartverifier/checks/charttesting.go
mmulholla: oops.

pkg/chartverifier/checks/charttesting.go
zonggen: What I propose was:
```
		if len(release) == 0 {
			release, namespace = chrt.CreateInstallParams(cfg.BuildId)
		} else {
			_, namespace = chrt.CreateInstallParams(cfg.BuildId)
		}
        if cfg.Namespace != "" {
            ...
        } else {
            ...
        }
```
`chrt.CreateInstallParams(cfg.BuildId)` was called at the begining of each block so not sure why they cannot be hoist up to the beginning to the function. But I think this is just personally preference, and definitely not a blocker.

mmulholla: my con with your code is that we may call chrt.CreateInstallParams(cfg.BuildId) when we don't need to because namespace was set in the cfg. Anyway we agree that there are a ton of ways to write this and all have pros ad cons. I kind of went with the minimum change on the current logic.

dperaza4dustbit (approved):
lgtm

pkg/chartverifier/checks/charttesting.go
mmulholla: also the existing code is well tested and has value based on how many times it has been run and worked. On this basis I try to not change more than I need to.

zonggen: agreed, resolving

zonggen (approved):

PR #220, opened 2021-12-01
add Chart Verifier acceptance test

mmulholla:
This PR covers: https://issues.redhat.com/browse/HELM-196

Adds a chart verifier acceptance test with 6 tests: chart src and chart tgz for partner, redhat and community profiles.

To run the tests manually just run "pytest" from the root directory of the repository. 

Some notes:
1. The tests are run in a separate workflow, at some point we need to decide if/how to incorporate into release flow.
2. The tests are run single threaded. Multithread does not work because the tests re-use the same chart which fails to install if already installed. This could be fixed by using --chart-set to set a value when the verifier is invoked and is used to set a variable which is used to set the app name for intsall. The test then ensures that value set is unique.
3. Each test includes a chart and an expected report of the the report.
    - good: Make it easy to add tests which can be defined by report of report constant - just update the feature file. 
    - bad: If report content changes the expected report of the report will need to be updated.

Here is a sample PR which should the new test running: https://github.com/mmulholla/chart-verifier/pull/94 (will not run on this PR)


scripts/src/buildandtest/buildandtest.py
```diff
@@ -103,24 +103,30 @@ def main():
...
+    parser.add_argument("-v", "--verifier-version", dest="verifier_version", type=str, required=False,
                         help="New version of chart verifier")
+    parser.add_argument("-b","--build-only",dest="build_only",type=str, required=False,
+                        help="set to any value to build only and not test")
 
 
     args = parser.parse_args()
-
-    image_id = args.image_name + ":" + args.sha_value[:7]
+    image_tag = args.sha_value[:7]
```
zonggen, line 113: I would suggest use a variable for '7' and give a proper name so later contributers do not need to guess the purpose of this value.

scripts/src/report/report_info.py
```diff
@@ -76,6 +82,13 @@ def get_report_chart(report_path):
      print("[INFO] report chart : %s" % metadata["chart"])
      return metadata["chart"]
 
+def get_all_reports(report_path,profile_type,profile_version):
+    all_reports = _get_report_info(report_path,ALL_REPORTS,profile_type,profile_version)
```
zonggen, line 86: nit: space separated parameters

tests/charts/psql-service/0.1.8/src/templates/deployment.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Values.k8Name }}
+  #namespace: {{ .Values.k8Project }}
```
zonggen, line 5:
```suggestion
```

tests/charts/psql-service/0.1.8/src/templates/deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+      annotations:
+        alpha.image.policy.openshift.io/resolve-names: '*'
+      labels:
+        app: {{ .Values.k8Name }}
+        phase: test
+    spec:
+      # securityContext:
+      #   runAsUser: {{ .Values.securityContext.runAsUser }}
+      #   fsGroup: {{ .Values.podSecurityContext.fsGroup }}
+      # serviceAccount: {{ .Values.serviceAccount }}
```
zonggen, line 24:
```suggestion
```

tests/charts/psql-service/0.1.8/src/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ .Values.k8ServiceName }}
+  #namespace: {{ .Values.k8Project }}
```
zonggen, line 5:
```suggestion
```

tests/charts/psql-service/0.1.8/src/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  type: {{ .Values.service.type }}
+  # loadBalancerSourceRanges:
+  #   - {{ .Values.service.sourceRange }}
```
zonggen, line 9:
```suggestion
```

tests/charts/psql-service/0.1.8/src/templates/tests/test-psql-connection.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ .Release.Name }}-credentials-test"
+  #namespace: {{ .Values.k8Project }}
```
zonggen, line 5:
```suggestion
```

tests/charts/psql-service/0.1.8/src/templates/tests/test-psql-connection.yaml
```diff
@@ -0,0 +1,26 @@
...
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  #serviceAccount: {{ .Values.serviceAccount }}
```
zonggen, line 9:
```suggestion
```

tests/charts/psql-service/0.1.8/src/values.schema.json
```diff
@@ -0,0 +1,59 @@
...
+            "pattern": "^[a-z0-9-_]+$"
+          },
+          "postgresDatabase": {
+            "type": "string",
+            "pattern": "^[a-z0-9-_]+$"
+          }
+        }
+      }     
+    }   
+  }
```
zonggen, line 59: nit: newline at EOF

zonggen (commented):
Overall LGTM, commented a few nits, great work!

tests/tests/functional/chart_test.py
```diff
@@ -0,0 +1,196 @@
...
+            local_chart = True
+        else:
+            return f"FAIL: chart does not exist: {os.path.abspath(chart_location)}"
+
+    if profile_type:
+        docker_command = docker_command + " --set profile.vendorType=" + profile_type
+
+    print(f'docker command: {docker_command}')
+
+    client = docker.from_env()
```
tisutisu, line 82: Can we can eliminate this? As "client" is already initialized in line no 43.

tests/tests/functional/chart_test.py
```diff
@@ -0,0 +1,196 @@
...
+def check_report(run_verifier, profile_type, report_info_location):
+
+    if run_verifier.startswith("FAIL"):
+        pytest.fail(f'FAIL some tests failed: {run_verifier}')
+
+    report_data = yaml.load(run_verifier, Loader=Loader)
+
+    report_vendor_type = report_data["metadata"]["tool"]["profile"]["VendorType"]
+    if report_vendor_type != profile_type:
+        print(f"FAIL: profiles do not match. Expected {profile_type}, but report has {report_vendor_type}")
```
tisutisu, line 120: Do we still want to continue if this step fails? I mean pytest.fail not needed here?

tisutisu (commented):
Overall LGTM. Have few comments, please do check them.

scripts/src/buildandtest/buildandtest.py
mmulholla: Good spot, I couldn't remember what the 7 was for so had to do some testing :-). It was to change the sha from long form to short form so after a bit of investigation I found a better way to do it. In build.yaml:  

```
commit_sha=$(git rev-parse --short HEAD)
ve1/bin/build-and-test --image-name=${{steps.check_version_in_PR.outputs.PR_release_image}} --sha-value=$commit_sha --verifier-version=${{steps.check_version_in_PR.outputs.PR_version}}
```

this way the 7 is no longer needed

scripts/src/report/report_info.py
mmulholla: nit fixed

tests/charts/psql-service/0.1.8/src/templates/deployment.yaml
mmulholla: For the changes in the chart - I just copied the chart from development repo, so I don't see any value in updating here until we have a common place to store charts.

tests/charts/psql-service/0.1.8/src/templates/deployment.yaml
mmulholla: see above.

tests/charts/psql-service/0.1.8/src/templates/service.yaml
mmulholla: see above

tests/charts/psql-service/0.1.8/src/templates/service.yaml
mmulholla: see above.

tests/charts/psql-service/0.1.8/src/templates/tests/test-psql-connection.yaml
mmulholla: see above

tests/charts/psql-service/0.1.8/src/templates/tests/test-psql-connection.yaml
mmulholla: see above.

tests/charts/psql-service/0.1.8/src/values.schema.json
mmulholla: see above.

mmulholla:
> Overall LGTM, commented a few nits, great work!

Thank you! Appreciate this comment.

tests/tests/functional/chart_test.py
mmulholla: eliminated, thanks.

tests/tests/functional/chart_test.py
mmulholla: good spot, we do want to continue to see other errors but set a boolean so test would fail even if this was the only failure.

mmulholla:
> Overall LGTM. Have few comments, please do check them.

Have addressed your comments, thanks for the review.

dperaza4dustbit (approved):
lgtm

PR #222, opened 2021-12-13
Add and test release tgz

mmulholla:
Build now creates and tests a tarball which includes:
- chart verifier executable.
- config directory containing profile definitions. 

Further if the PR is for a release, the tested tarball is added as a release asset:

Some notes:
- Reduced the workflow to a single workflow (merged acceptance test into Build.yaml)
- Modified acceptance tests to run the same tests on both the docker image and the chart verifier executable extracted from the tarball.
- The tarball is created and tested even if the PR is not for a release.
- The name of the tarball does not include the os it runs on. Maybe should change this. 

The build for this PR will fail because it will not use the updated Build.yaml file. 

An example run for a release is shown here: https://github.com/mmulholla/chart-verifier/pull/109
And the release it created: https://github.com/mmulholla/chart-verifier/releases/tag/1.4.1


tests/tests/functional/chart_test.py
```diff
@@ -105,6 +107,18 @@ def run_verifier(image, profile_type, chart_location):
...
     return output.decode("utf-8")
 
+def run_tarball_image(tarball_name,profile_type, chart_location):
+    print(f"Run tarball image from {tarball_name}")
+
+    tar = tarfile.open(tarball_name, "r:gz")
+
+    tar.extractall(path="./test_verifier")
+
+    out = subprocess.run(["./test_verifier/chart-verifier","verify","--set",f"profile.vendorType={profile_type}",chart_location],capture_output=True)
```
dperaza4dustbit, line 117: So much simpler the tar test.

dperaza4dustbit: Curious... how come this works without PR 221 merged. In the Docker Images we already have, helm, oc and kubectl binaries, but don't see here where they get added.

dperaza4dustbit (commented):
Looks good, just one question.

tests/tests/functional/chart_test.py
dperaza4dustbit: In the test I see is not getting this far as it fails when tart is being opened.

mmulholla: The test fails because it uses the build.yaml file from the main branch which does not include the changes to create the tarball. I did include a link to a PR which uses the new build.yaml. Guess the tarball test passes because helm and oc are installed.

dperaza4dustbit: Got it.

dperaza4dustbit (approved):
lgtm

zonggen (approved):
LGTM

PR #233, opened 2022-01-11
create verifier tarball for all builds

mmulholla:
PR test failing for non release builds because verifier tarball was not being created. Update the code to always generate the tarball.

### DIFF
--- a/scripts/src/release/releasechecker.py
+++ b/scripts/src/release/releasechecker.py
@@ -87,18 +87,19 @@ def main():
                         help="Version to compare")
 
     args = parser.parse_args()
-    if args.api_url and check_if_only_version_file_is_modified(args.api_url):
-        ## should be on PR branch
+    if args.api_url:
         version_info = get_version_info()
-        print(f'[INFO] Release found in PR files : {version_info["version"]}.')
-        print(f'::set-output name=PR_version::{version_info["version"]}')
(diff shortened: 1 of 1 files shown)


scripts/src/release/releasechecker.py
```diff
@@ -87,18 +87,19 @@ def main():
...
-        print(f'::set-output name=PR_release_info::{version_info["release-info"]}')
-        print(f'::set-output name=PR_includes_release::true')
-        make_release_body(version_info["version"],version_info["quay-image"],version_info["release-info"])
         asset_file = tarfile_asset.create(version_info["version"])
-        print(f'[INFO] Release asset created : {asset_file}.')
+        print(f'[INFO] Verifier tarball created : {asset_file}.')
         print(f'::set-output name=PR_tarball_name::{asset_file}')
+        if check_if_only_version_file_is_modified(args.api_url):
+            ## should be on PR branch
+            print(f'[INFO] Release found in PR files : {version_info["version"]}.')
```
zonggen, line 97: `version_info[key]` will raise key error if the key does not exist, it's safer to use `version_info.get(key,default_value)` to avoid key errors. But it also depends on whether we want to produce the error if the key does not exist.

mmulholla: It looks like I added that code but i did not, it was there before so I propose leaving as-is.

zonggen (approved):

Issue #224, opened 2021-12-14
Panic running version1.4.0 and newer


betoredhat:
Panic on version 1.4.1

```
cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="8.5 (Ootpa)"
```

```
 podman info
host:
  arch: amd64
  buildahVersion: 1.22.3
  cgroupControllers: []
  cgroupManager: cgroupfs
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.29-1.module+el8.5.0+12582+56d94c81.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.29, commit: 0f5bee61b18d4581668e5bf18b910cda3cff5081'
  cpus: 80
  distribution:
    distribution: '"rhel"'
    version: "8.5"
  eventLogger: file
  hostname: jumphost.dfwt5g.lab
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1004
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1004
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 4.18.0-305.3.1.el8_4.x86_64
  linkmode: dynamic
  memFree: 15542751232
  memTotal: 540411604992
  ociRuntime:
    name: runc
    package: runc-1.0.2-1.module+el8.5.0+12582+56d94c81.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.2
      spec: 1.0.2-dev
      go: go1.16.7
      libseccomp: 2.5.1
  os: linux
  remoteSocket:
    path: /run/user/1004/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_NET_RAW,CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.1.8-1.module+el8.5.0+12582+56d94c81.x86_64
[...]

PR #237, opened 2022-01-13
Fix podman error and add tests

mmulholla:
Fixed the reported panic error and added tests.

Note: the podman tests will fail in this build because it is using the old build.yml file which does not build the podman image. As a result the test uses the main image from quay which is not fixed. 

This PR on my branch is using the updated build.yml and  shows the tests passing: https://github.com/mmulholla/chart-verifier/pull/123

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -25,7 +25,7 @@ jobs:
               uses: actions/setup-go@v2
               with:
                 go-version: '1.17.2'
-
+                
             - name: Check go mod status
               working-directory: ./chart-verifier
               run: |
@@ -117,6 +117,19 @@ jobs:
                   commit_sha=$(git rev-parse --short HEAD)
                   ve1/bin/build-and-test --image-name="quay.io/redhat-certification/chart-verifier" --sha-value=$commit_sha --build-only="True"}
 
+
+            - name: Build podman Image
+              working-directory: ./chart-verifier
+              id: build_podman_image
+              run: |
+                  # build a podman image
+                  # sudo apt install qemu-system-x86
+                  commit_sha=$(git rev-parse --short HEAD)
+                  image_tag="podman-"$commit_sha
+                  echo "use image tag $image_tag"
+                  podman build -t quay.io/redhat-certification/chart-verifier:$image_tag .
+                  echo "::set-output name=podman_image_tag::$image_tag"
+
             - name: Login to oc
               working-directory: ./chart-verifier
               env:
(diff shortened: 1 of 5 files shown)


zonggen (commented):
This PR LGTM, however it might be more flexible if we could set this config dir as a parameter instead of hard-coded (e.g. through a flag or env variable or both). WDYT?

The reason I mentioned this is because I was facing the same error when using cv binary (see https://github.com/redhat-certification/chart-verifier/issues/224#issuecomment-1006071795).

mmulholla:
> This PR LGTM, however it might be more flexible if we could set this config dir as a parameter instead of hard-coded (e.g. through a flag or env variable or both). WDYT?
> 
> The reason I mentioned this is because I was facing the same error when using cv binary (see [#224 (comment)](https://github.com/redhat-certification/chart-verifier/issues/224#issuecomment-1006071795)).

@zonggen  Not sure I understand what you are suggesting, could you elaborate a bit please.

zonggen:
Sorry about the confusion, I am assuming currently `chart-verifier` is looking through pre-determined location to get profiles (https://github.com/redhat-certification/chart-verifier/blob/main/pkg/chartverifier/profiles/profile.go#L119). If we could add a flag to pass in the path `--profile-dir` or `--profile`, or an environment variable like `PROFILE_PATH`, it would mitigate similar issues where `chart-verifier` cannot find the profiles. Because of the issue, `chart-verifier` binary does not work currently because it cannot find the config locally.

mmulholla:
> Sorry about the confusion, I am assuming currently `chart-verifier` is looking through pre-determined location to get profiles (https://github.com/redhat-certification/chart-verifier/blob/main/pkg/chartverifier/profiles/profile.go#L119). If we could add a flag to pass in the path `--profile-dir` or `--profile`, or an environment variable like `PROFILE_PATH`, it would mitigate similar issues where `chart-verifier` cannot find the profiles. Because of the issue, `chart-verifier` binary does not work currently because it cannot find the profiles locally.
> 
> cc @divyansh42

@zonggen , thank you for the explanation.

For the docker and podman images we include the profiles in the images in a standard place and the failure to find them by the podman image was a bug which this PR fixes. I don't think a flag would help here. 

For the binary we include the profiles in the binary - in a config dir and assume that you run the chart verifier from the directory you unzipped it into, otherwise we fail to find the profiles. We can discuss how to improve this  in a scrum and definitely fail in a better was than we currently do, but I think this would be separate from this PR which is for the podman case.

zonggen:
I don't have objections in merging this PR, and think the flag would help in this issue as well since we don't need to hardcode the path like https://github.com/redhat-certification/chart-verifier/pull/237/files#diff-4729c2745c355e487c5eb4a3d8868254f745cf832ffb15c5c05e99f068ef4716R123. So within a container runtime (either docker or podman or any other runtime), we would mount the directory with profiles and modify the entrypoint to find the profiles by using a flag or environment variable: https://github.com/redhat-certification/chart-verifier/blob/main/Dockerfile#L36. As in the binary case, I tried to move the binary into the config dir but still I saw the same error.

zonggen (approved):

PR #238, opened 2022-01-19
embed profiles in images

mmulholla:
Have redone the fix to

- move the profiles into pkg/profileconfig/profiles
- add pkg/profileconfig/profiles.go to provide a method to get the content of the profiles.
- modify pkg/chartverifier/profiles/profile.go to use the new method in pkg/profileconfig/profiles.go to get the profiles.
- modify cmd/version.go to embed the cmd/release/release_info.json file.
- modify Dockerfile to not copy release_info.json file and profiles into image folder.
- modify release checker to not include the profiles in the tarball.
- add a fix to prevent the panic error (which should never happen now)


Makefile
```diff
@@ -14,15 +14,21 @@ gofmt:
 
 .PHONY: bin
 bin:
-	 go build -o ./out/chart-verifier main.go
+	cp -r ./config ./pkg/chartverifier/profiles
+	go build -o ./out/chart-verifier main.go
+	rm -rf ./pkg/chartverifier/profiles/config
```
isutton, line 19: Perhaps a new target can be created

```Makefile
.PHONY: copy-config
copy-config:
        rm -fr ./pkg/chartverifier/profiles/config
        cp -r ./config ./pkg/chartverifier/profiles

.PHONY: bin
bin: copy-config
        go build -o ./out/chart-verifier main.go
```

Makefile
```diff
@@ -14,15 +14,21 @@ gofmt:
...
 
 .PHONY: bin_win
 bin_win:
+	cp -r ./config ./pkg/chartverifier/profiles
 	env GOOS=windows GOARCH=amd64 go build -o .\out\chart-verifier.exe main.go
+	rm -rf ./pkg/chartverifier/profiles/config
```
isutton, line 25:
```suggestion
bin_win: copy-config
	env GOOS=windows GOARCH=amd64 go build -o .\out\chart-verifier.exe main.go
```

Makefile
```diff
@@ -14,15 +14,21 @@ gofmt:
...
 
 .PHONY: test
 test:
+	cp -r ./config ./pkg/chartverifier/profiles
 	go test -v ./...
+	rm -rf ./pkg/chartverifier/profiles/config
```
isutton, line 31:
```suggestion
test: copy-config
	go test -v ./...
```

mmulholla:
Have redone the fix to 
- move the profiles into pkg/profileconfig/profiles
- add pkg/profileconfig/profiles.go to provide a method to get the content of the profiles.
- modify pkg/chartverifier/profiles/profile.go to use the new method in pkg/profileconfig/profiles.go to get the profiles.
- modify cmd/version.go to embed the cmd/release/release_info.json file.
- modify Dockerfile to not copy release_info.json file and profiles into image folder.

Makefile
mmulholla: decided to move profiles to a directory under pkg so this change is no longer needed.

Makefile
mmulholla: decided to move profiles to a directory under pkg so this change is no longer needed.

Makefile
mmulholla: decided to move profiles to a directory under pkg so this change is no longer needed.

zonggen:
This seems to be requing `glibc-2.34` and above. I compiled the binary in my host machine and run it in a virtual machine, got the following:
```bash
$ ./chart-verifier 
./chart-verifier: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./chart-verifier)
```

zonggen (approved):

dperaza4dustbit (approved):
lgtm, I would still leave the code around to load the files as this will be needed when we add the option to override the profile.

PR #240, opened 2022-01-27
 report to stdout and error log

mmulholla:
New behavior:

Default:

- report goes to stdout
- all other messages go to stderr
- all other messages and log are written to chartverifier/verifier-```<timestamp>```.log
   -  Once 10 files have been created, the oldest will be deleted before a new one is created.

Options

- write report to a file. e.g.: chartverifier/report.yaml
   - if file exists is is replaced.
- suppress log file output
- errors are still output to stderr.

Notes:

 - log-ouput - flag is removed

Have also added a new unit test and updated other test for report output to stdout. 

Will update README separately to time with a release.


dperaza4dustbit (approved):
lgtm

PR #241, opened 2022-01-31
update docs to report to stdout and error log

mmulholla:
Update verify help and readme as a result of changing report output to stdout and adding an error log


dperaza4dustbit (approved):
lgtm

PR #246, opened 2022-03-01
add 30 minute timeout for helm test

mmulholla:
Prevent hangs if tests take too long - set timeout to 30 minutes to leave plenty of time.
Also improved chart install and testing check reason to include if the failure was in install or test.
for example:
```
- check: v1.0/chart-testing
  type: Mandatory
  outcome: FAIL
  reason: 'Chart Install failure: timed out waiting for the condition'
```

### DIFF
--- a/pkg/tool/helm.go
+++ b/pkg/tool/helm.go
@@ -102,8 +102,9 @@ func (h Helm) Test(namespace, release string) error {
 	utils.LogInfo(fmt.Sprintf("Execute helm test. namespace: %s, release: %s, args: %+v", namespace, release, h.args))
 	client := action.NewReleaseTesting(h.config)
 	client.Namespace = namespace
+	client.Timeout = 30 * time.Minute
 
-	// TODO: support filter and timeout options if required
+	// TODO: support filter
 	_, err := client.Run(release)
 	if err != nil {
 		utils.LogError(fmt.Sprintf("Execute helm test. error %v", err))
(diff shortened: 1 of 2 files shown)


pkg/tool/helm.go
```diff
@@ -102,8 +102,9 @@ func (h Helm) Test(namespace, release string) error {
 	utils.LogInfo(fmt.Sprintf("Execute helm test. namespace: %s, release: %s, args: %+v", namespace, release, h.args))
 	client := action.NewReleaseTesting(h.config)
 	client.Namespace = namespace
+	client.Timeout = 30 * time.Minute
```
dperaza4dustbit, line 105: Right now seems to be timing out at 5 minutes, is the intention to give it more time to test? This will delay our checks by quite a bit on some cases, do you think this will impact our test harness on certification flow?

dperaza4dustbit (commented):
quick question

pkg/tool/helm.go
mmulholla: Yes the timeout is set to prevent a false failure if the test took between 5-30 minutes. Difficult to know how long a test will take but for 99% of PR's I suspect the timeout will be irrelevant because they will complete or fail fast, just gives a test plenty of time to complete before we think it is a hang. Can reduce if you prefer.

dperaza4dustbit: I think is a good start, thanks for explaining

dperaza4dustbit (approved):
lgtm

PR #249, opened 2022-03-02
create dev release after merge

mmulholla:
Creates a new dev release each time a PR is merged into the main branch. The release includes a binary which can be consumed by the chart verifier github action.

Notes: 
- regular PR's are not auto-merged so creating the release could not be part of the regular workflow.
- i could add a slack message each rime a release is created if needed.
- the first dev release will be created when this PR is merged - causes the new action to run.


dperaza4dustbit (approved):
lgtm

PR #258, opened 2022-04-21
allow for long line in template file

mmulholla:


pkg/chartverifier/checks/helm.go
```diff
@@ -200,27 +201,39 @@ func getImageReferences(chartUri string, vals map[string]interface{}) ([]string,
 	mem.SetNamespace("TestNamespace")
 	actionConfig.Releases = storage.Init(mem)
 
-	imagesMap := make(map[string]bool)
-
 	txt, err := actions.RenderManifests("testRelease", chartUri, vals, actionConfig)
+	if err != nil {
+		return nil, err
```
isutton, line 206: Wrap the error, so we know where the error comes from:

```go
return nil, fmt.Errorf("error rendering manifests: %w", err)
```

Tagging the error allow us to easily find the source of issue in a debugging session.

pkg/chartverifier/checks/helm.go
```diff
@@ -200,27 +201,39 @@ func getImageReferences(chartUri string, vals map[string]interface{}) ([]string,
...
+
+}
+
+func getImagesFromContent(content string) ([]string, error) {
+
+	imagesMap := make(map[string]bool)
 
 	type ImageRef struct {
 		Ref string `yaml:"image"`
 	}
```
isutton, line 219: Any particular reason this type is defined in the context of the enclosing function? I assume its content is serialized somewhere, as the `yaml` tag is being explicitly used.

Unless major changes are required, I'd also move this type outside the enclosing function.

pkg/chartverifier/checks/helm.go
```diff
@@ -200,27 +201,39 @@ func getImageReferences(chartUri string, vals map[string]interface{}) ([]string,
...
+			if len(imageRef.Ref) > 0 && !imagesMap[imageRef.Ref] {
+				imagesMap[imageRef.Ref] = true
 			}
 		}
+		line, err = getNextLine(reader)
+		if err == io.EOF {
+			err = nil
+			break
+		}
 	}
```
isutton, line 237: I would rewrite this do-while pattern as the following suggestion:

```suggestion
	for line, err := getNextLine(reader); err != io.EOF {
		if err != nil {
			// an error occured
			break
		} 
		var imageRef ImageRef
		yamlErr := yaml.Unmarshal([]byte(line), &imageRef)
		if yamlErr == nil {
			if len(imageRef.Ref) > 0 && !imagesMap[imageRef.Ref] {
				imagesMap[imageRef.Ref] = true
			}
		}
	}
```

pkg/chartverifier/checks/helm.go
```diff
@@ -229,4 +242,17 @@ func getImageReferences(chartUri string, vals map[string]interface{}) ([]string,
...
+func getNextLine(reader *bufio.Reader) (string, error) {
+	nextLine, isPrefix, err := reader.ReadLine()
+	if isPrefix && err == nil {
+		for isPrefix && err == nil {
+			var partLine []byte
+			partLine, isPrefix, err = reader.ReadLine()
+			nextLine = append(nextLine, partLine...)
+		}
+	}
+	return string(nextLine), err
```
isutton, line 257: bufio.Reader already has a similar function:

```suggestion
	return reader.ReadString("\n")
```

Perhaps this function should be removed and bufio.Reader.ReadString() used in place?

isutton (changes requested):
There are some improvements still to be made.

pkg/chartverifier/checks/helm.go
mmulholla: Not part is his fix so will leave as-is

pkg/chartverifier/checks/helm.go
mmulholla: You loop has issues, the scope of err is limited to the loop so is not available for return, then if I change the scope of err the loop wold exit with err set to io.EOF so would need setting back to nil.  As a result I left loop as-is as it was the simplest thing to do and I know it works.

isutton: Let's then rename `getNextLine` error to `gnlErr`, and copy it to the function scoped `err` before break:

```go
// gnlErr will be discarded and the resolution error won't leak to `err`
for line, gnlErr := getNextLine(reader); gnlErr != io.EOF {
        if gnlErr != nil {
                err = gnlErr  // assign the local error to the function scoped `err`
                break
        }
        ...
}
```

isutton (approved):

Issue #259, opened 2022-04-29
Missing tags from repositories with many tags


ebaron:
I have a test that uses the `ubi8/ubi` image. I've tried with both `latest` and `8.5`, but in both cases I get the result:
`
[...]

PR #265, opened 2022-06-02
allow for pyxis responses spanning more than one page

mmulholla:
Updated code to allow for pyxis returning data in multiple pages. Also added some logging to image certification check. 
For testing added test for ```registry.access.redhat.com/rhscl/postgresql-10-rhel7:latest``` which is only found on the second pyxis page. 

see: https://github.com/redhat-certification/chart-verifier/issues/259
see: https://issues.redhat.com/browse/HELM-361

Added a second commit or a second fix needed by IBM. Basically now get the image sha from  the ```image_id```  attribute instead of ```parsed_data.docker_image_id``` which seemed to disappear. The new attribute was recommended by the Pyxis teaam.
see: https://issues.redhat.com/browse/HELM-362

### DIFF
--- a/pkg/chartverifier/pyxis/pyxis.go
+++ b/pkg/chartverifier/pyxis/pyxis.go
@@ -22,12 +22,17 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+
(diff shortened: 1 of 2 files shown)


Kartikey-star (approved):
lgtm

PR #270, opened 2022-06-29
Add note about support case requirement to set providerDelivery flag in OWNERS file.

jfrancin:
There are presently no breadcrumbs in the docs to tell partners that adding the providerDelivery flag to the OWNERS file can only be done by Engineering via a support case. Add a note to that effect.


mmulholla:
No longer needed. Now added as part of OWNER file creation,

PR #272, opened 2022-07-25
move kube-ocp version map to yaml file for wider use

mmulholla:
I made this change as part of the work for https://issues.redhat.com/browse/HELM-358.

For helm 389 the latest kubeversion-ocp version map is needed, so rather than create it in two places, where one gets out of date, it would be better to read the yaml file from the chartverifier repo which should always contain the latest info.

Also added a quick fix for a test which failed after the new api was merged.


internal/tool/kubectl.go
```diff
@@ -12,20 +14,48 @@ import (
...
+
+type versionMapping struct {
+	KubeVersion string `yaml:"kube-version"`
+	OcpVersion  string `yaml:"ocp-version"`
+}
+
+func init() {
+	kubeOpenShiftVersionMap = make(map[string]string)
+
+	yamlFile, err := content.ReadFile("kubeOpenShiftVersionMap.yaml")
```
Kartikey-star, line 39: I would suggest the flow should be like this:

	
yamlFile, err := content.ReadFile("kubeOpenShiftVersionMap.yaml")
if err!= nil {
		utils.LogError(fmt.Sprintf("Error reading content of kubeOpenShiftVersionMap.yaml: %v", err))
}

versions := versionMap{}

err = yaml.Unmarshal(yamlFile, &versions)

if err != nil {
utils.LogError(fmt.Sprintf("Unmarshall error reading kubeOpenShiftVersionMap.yaml: %v", err))
}

for _, versionMap := range versions.Versions {
	kubeOpenShiftVersionMap[versionMap.KubeVersion] = versionMap.OcpVersion
}

Having an if/for within a for should be avoided in my opinion.

Kartikey-star (commented):
Looks good to me. Have a comment. Please take a look on kubectl.go init function.

internal/tool/kubectl.go
mmulholla: Single exit code has its benefits, and code review should not be about opinions.  Updated it anyway.

Kartikey-star (approved):

PR #277, opened 2022-08-26
SA token creation fix in 4.11

tisutisu:
This PR fixes the sa token creation issue in 4.11 during chart-testing


mmulholla (approved):
LGTM

PR #281, opened 2022-09-22
add new signed chart and set main version to 1.9.0 for testing

mmulholla:
Added a signed chart ready for testing
Update version to 1.9.0 in main for testsing.

Note - adding chart to PR prevent version update from creating a release.


tisutisu (approved):
LGTM

PR #282, opened 2022-09-29
Bump versions helm==3.9.4 and mako==1.2.3 : Dependabot Alert

tisutisu:
It will address dependabot alerts here https://github.com/redhat-certification/chart-verifier/security/dependabot


mmulholla (approved):
LGTM

PR #287, opened 2022-10-19
add signed chart check, tests and doc

mmulholla:
- Added new check for signed charts
- Added unit tests
- Added python test

Chart verifier logic for new check is documented here: https://lucid.app/lucidchart/1fd9bb6c-98b9-4a20-8d69-8af003ce8ee1/edit?invitationId=inv_99e435e0-1db6-4181-873d-4ce60bef8d48&page=0_0#

Jora issue: https://issues.redhat.com/browse/HELM-410


cmd/verify.go
```diff
@@ -238,9 +249,47 @@ func NewVerifyCmd(config *viper.Viper) *cobra.Command {
...
 	cmd.Flags().BoolVarP(&providerDelivery, "provider-delivery", "d", false, "chart provider will provide the chart delivery mechanism (default: false)")
+	cmd.Flags().StringVarP(&pgpPublicKeyFile, "pgp-public-key", "k", "", "file containing gpg public key of the key used to sign the chart")
 	return cmd
 }
 
 func init() {
 	rootCmd.AddCommand(NewVerifyCmd(viper.GetViper()))
 }
+
+func readPublicKeyFile(keyFilename string) (string, error) {
```
Kartikey-star, line 260: Is the function readPublicKeyFile getting used somewhere.If not then we can remove this.

internal/chartverifier/verifier.go
```diff
@@ -115,13 +118,23 @@ func (c *verifier) Verify(uri string) (*apiReport.Report, error) {
...
+			PublicKeys:       c.publicKeys,
 		})
 
 		if checkErr != nil {
 			return nil, NewCheckErr(checkErr)
 		}
 		_ = result.AddCheck(check, r)
 
+		if check.CheckId.Name == apiChecks.SignatureIsValid {
+			if len(c.publicKeys) == 1 && strings.Contains(r.Reason, checks.ChartSigned) {
```
Kartikey-star, line 130: The public keys is an array of string, but here we are checking whether it contains only one public key in c.publicKeys array. So if the user has to specify the public key per chart then why not have the type of c.publicKeys as a string instead of array.

internal/chartverifier/checks/helm.go
```diff
@@ -119,21 +119,18 @@ func (c *chartCache) Get(uri string) (ChartCacheItem, bool, error) {
...
-			return ChartCacheItem{}, err
-		}
-	} else {
-		userCacheDir = repositoryCache
+
+	userCacheDir = getCacheDir(opts)
+	if userCacheDir == "" {
+		return ChartCacheItem{}, err
 	}
 	cacheDir := path.Join(userCacheDir, "chart-verifier")
```
Kartikey-star, line 132: If this line `cacheDir := path.Join(userCacheDir, "chart-verifier")` required here now? getCacheDir(opts *CheckOptions) func seems to be returning the repocache/usercache with chart-verifier joined to the path.

cmd/verify.go
mmulholla: Removed it

internal/chartverifier/verifier.go
mmulholla: With an array we do not know which key was the correct one so cannot add digest to report. With cli we only get one key so if chart verifies we know that is the correct key so we can then add digest to report.

internal/chartverifier/checks/helm.go
mmulholla: Goof spot, no longer needed. Removed.

Kartikey-star (approved):
Looks good to me .As discussed we would need to change the way the public keys of developer are used for testing purpose but that should be addressed with a different PR.

PR #290, opened 2022-11-30
Check release exists

mmulholla:
For a release, if the version before and after are the same, still create a release if the releases does not already exist.

Also update tests to allow for a certified image tests use being removed from certified images in pyxis.


PatAKnight:
LGTM

PatAKnight (approved):

PR #292, opened 2022-12-01
Fix issue where pyxis image lookups fail for RH TBR

jsm84:
Summary:
* Fixes an issue where Red Hat provided images fail pyxis API lookup if hosted in the Terms Based Registry (TBR)

Signed-off-by: Josh Manning <19478595+jsm84@users.noreply.github.com>

### DIFF
--- a/internal/chartverifier/pyxis/pyxis.go
+++ b/internal/chartverifier/pyxis/pyxis.go
@@ -148,6 +148,10 @@ Loops:
 		nextPage := 0
 		allDataRead := false
 
+		// All Red Hat images are stored in pyxis under the registry.access.redhat.com URL
+		if registry == "registry.redhat.io" {
+			registry = "registry.access.redhat.com"
+		}
 		requestUrl := fmt.Sprintf("%s/registry/%s/repository/%s/images", pyxisBaseUrl, registry, imageRef.Repository)
 		utils.LogInfo(fmt.Sprintf("Search url: %s, tag: %s, sha: %s ", requestUrl, imageRef.Tag, imageRef.Sha))
 


jsm84:
@mmulholla I've rebased this against the `go test` changes in main, and this is ready for approval and merge.

mmulholla:
@jsm84 we have been working on a change where we skip images with a registry of registry.redhat.com - are you suggesting from this that all images with that registry are in fact certified but just under the different registry. Also would like to see a quick go test added.

mmulholla:
@jsm84 Another question will the image still be available in registry.redhat.io or should the chart really be updated also? It would be wrong to certify a chart by making the change when it chart itself will still would not work.

jsm84:
@mmulholla To answer your first question, yes, regardless of whether an image is only available through registry.redhat.io or not, all Red Hat provided container images are stored in pyxis under the registry.access.redhat.com URL. This was clarified to me by Jakub Paulovic, internally.

As for your second question, the images used in Yugabyte's most recent chart (which spurred this fix) are only available through registry.redhat.io (as terms acceptance is enforced through the TBR), and not the usual registry.access.redhat.com (which is unauthenticaed).

This created a rock/hard-place scenario where the partner was unable to change their helm chart to reflect how the image was stored in pyxis under registry.access.redhat.com, because two images (openshift4/ose-prometheus and rhscl/postgresql-13-rhel7).

The simplest fix here was to simply adjust the registry used for the pyxis image lookup (only in the case that the registry matches `registry.redhat.io`), but leaving the image pull specs from the chart untouched.

LMK if that makes sense. I'm happy to submit a new go test case, also.

mmulholla:
@jsm84 Thanks for the update. 

We discussed and would like a more complex fix:

1. Check if registry.redhat.io image is certified using lookup in pyxis (would this ever work?)
2. If not certified switch to registry.access.redhat.com and try again. 
     -  If image is found: 
         - Get sha for image for pyxis data and compare with image in registry.redhat.io. (us there a way to get sha for registry.redhat.io image without a userid and password?)
             - if sha's match image is certified - pass
             - if sha's do not match image is not certified - fail.
             - if cannot get sha for images image certification - skipped (treated as a pass)
    - If image is not found it is not certified- fail.

The thing we are trying to avoid is changing the registry from registry.redhat.io to registry.access.redhat.com and assuming they refer to exactly the same image.

The other interesting thing about the yugabyte chart is that this iamge: registry.access.redhat.com/rhscl/postgresql-13-rhel7:1-88.16615317220 does not exist in pyxis - that tag is not found.

jsm84:
@mmulholla That makes sense, in that you want to make sure the image is exactly the same between the two registries.

I do have some further input on a few of your points:

1. AFAIK there are zero images stored in pyxis under the registry.redhat.io registry. I could do some further digging to confirm. But if what I was told is true (and thus far does appear to be), then this suggestion would fail 100% of the time.

2. Even though the image check would always fail in such case, I think it makes sense to check that the digest matches between the two registries. If the image is the same (sha256 digest matches), and exists in pyxis under the registry.access.redhat.com registry (which it always should), then you know that the image is indeed certified.

 There are a couple of caveats to this approach:

1. The terms based registry (TBR) aka registry.redhat.io enforces a valid RH username and password. You can't even read the manifest to simply check the image digest, if unauthenticated. We could add a feature for the user to supply a pull secret (the same pull secret used to install OpenShift works to this end) to authenticate to the TBR.

2. This is not really a caveat, but I highly suspect that the TBR is just a front-end proxy of sorts for registry.access.redhat.com, that is able to exclusively serve content which is marked to require such terms acceptance. This is probably why we store RH image data in pyxis under registry.access.redhat.com, regardless. As such, I would always expect a digest check to match (but it still doesn't hurt).

As for the Yugabyte chart, I think the 1-88.xxxx tag had dropped off (as in, I suspect it was removed by us/RH) in the time since they were granted an exception on their chart. This image came back as certified in previous `chart-verifier` runs.

Thank you for taking the time to discuss. I'm OK with whatever the decision is, but it's best to avoid any extra code that ends up being a NoOp.

mmulholla:
Closing this PR as no longer needed: See:https://github.com/redhat-certification/chart-verifier/pull/293

PR #293, opened 2022-12-02
update images are certified check to skip registry.redhat.io images

PatAKnight:
Jira: https://issues.redhat.com/browse/HELM-439

Adds a new v1.1/images-are-certified check to the 1.2 profiles.

Originally, registry.redhat.io images would fail during the images-are-certified check because they were not available through pyxis. With the update, the v1.1/images-are-certified check will now mark registry.redhat.io images as skipped.

Added the a new unit test for the updated v1.1/images-are-certified check with test cases that include an images-are-certified passing chart that contains skipped images, an images-are-certified failing chart that contains skipped images, and an images-are-certified skipped chart that only contains skipped images.


docs/helm-chart-troubleshooting.md
```diff
@@ -104,7 +104,31 @@ Requires any images referenced in a chart to be Red Hat Certified.
       match is not found the check fails.
     - if the image specified a tag value it is compared with the `repositories.tags.name` attributes. If a match is
       not found the check fails.
-- If the check fails use the point fo failure to determine how to address the issue. 
+- If the check fails use the point of failure to determine how to address the issue. 
+
+### `images-are-certified` v1.1
```
mmulholla, line 109: I would just do this as this is what is changed in v.1.1 rather than repeat the entire section.

mmulholla (approved):
LGTM

PR #294, opened 2022-12-06
Update provider deliver flag to web catalog only

PatAKnight:
Jira: https://issues.redhat.com/browse/HELM-461

Updates the provider delivery flag to instead say web catalog only.

Mentions of provider delivery have led to some confusions with users. This is due to the fact that the [Creating a Helm Chart Certification Project](https://redhat-connect.gitbook.io/partner-guide-for-red-hat-openshift-and-container/helm-chart-certification/creating-a-helm-chart-certification-project) article does not have any mentions of a provider delivery distribution method. In our case provider delivery is the same as the web catalog only distribution method.

Also updated the unit test that utilizes the provider delivery flag to now match the updates from this PR.

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -30,6 +30,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/cli/values"
+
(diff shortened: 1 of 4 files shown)


mmulholla (approved):
LGTM

PR #297, opened 2022-12-14
[fix] equal sign lost in chart set values

ethinx:
When there are equal sign `=` in chart set values, they may be removed before passing to helm. 

e.g.

```
./chart-verifier verify -S config.key='ABC/==' ....
```

The last two `=` will not be passed to helm.


jsm84 (approved):
Reviewed and I agree that base64 values could be affected here. I've tested the key parts of this in go playground to confirm.

LGTM

komish:
I have rebased your PR and have it running through tests. After that, we'll get it merged. Thanks for your contribution.

PR #298, opened 2022-12-14
increase timeout waitining for quay image to appear during release

mmulholla:
During the release process the workflow waits 15 minutes for the quay image to build and be ready. Recently this timeout is always popping as quay seems to be taking longer and this adds a manual step the release process which can be error prone, Increasing the timeout to


PatAKnight (approved):
LGTM

mmulholla:
Build failure is a Pyxis communication issue and not caused by this fix.

PR #301, opened 2022-12-15
Deprecation of set-output

PatAKnight:
Jira: https://issues.redhat.com/browse/HELM-438

Removes the deprecated set-output in favor for GITHUB_OUTPUT

The workflows and associated scripts make extensive use of set-output, but it is now being deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This PR removes set-output and replaces it with GITHUB_OUTPUT within the workflow. It also introduces a utils module that will handle adding the output to GITHUB_OUTPUT for the python scripts.

### DIFF
--- a/.github/workflows/Build.yaml
+++ b/.github/workflows/Build.yaml
@@ -92,7 +92,7 @@ jobs:
             - name: Get Date
               id: get-date
               run: |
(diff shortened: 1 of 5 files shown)


mmulholla (approved):
LGTM

PR #303, opened 2022-12-15
Fix error in link image script

mmulholla:
In the release image tag linking was not working because the response fro pyxis no longer includes an ```image_id``` attribute. This cause an exception and the retry logic kicks in. Have changed this to use ```manifest_digest``` instead. Ran the script locally to ensure it works.


PatAKnight (approved):
LGTM

PR #305, opened 2023-01-12
change report to webCatalogOnly

mmulholla:
Updated code to use web catalog only as a term of choice and to be set in the report.


docs/helm-chart-submission.md
```diff
@@ -45,7 +45,9 @@ For more information on the different Helm Chart Distribution methods, see: [Cre
 
 ## Web catalog only delivery
 
-Web catalog only delivery was previously described as provider delivery. Chart submissions will be made available within the Helm Chart Catalog on successful certifications. In the event that this case is undisirable, the provider should consider the web catalog only option if they wish not to make the chart publicly available. With web catalog only delivery the provider of the chart controls access to the chart and this will impact report generation:
+Web catalog only delivery was previously described as provider delivery. 
+
+Chart submissions will be made available within the Helm Chart Catalog on successful certifications. In the event that this case is undesirable, the provider should consider the web catalog only option if they wish not to make the chart publicly available. With web catalog only delivery the provider of the chart controls access to the chart and this will impact report generation:
```
Kartikey-star, line 50: Is the last sentence meant to read like `With web catalog only delivery the provider of the chart controls access to the chart and this will impact report generation:`

Kartikey-star (approved):
Looks good to me. May be we can think of an alternative to `web catalog only delivery` usage in the docs. Otherwise the changes looks good.

PR #312, opened 2023-02-10
Ensure the chart's verifier-version matches the verifier version subcommand output

komish:
The chart-verifier container image 1.8.0 would produce reports that reported a `.metadata.tool.verifier-version` of 1.0.0.

This PR adds an extension of the BDD that would check the charts produced against the version output of the built binary/image's `version` subcommand to make sure that they align.

The original scoped implementation suggested doing this at the point where a release asset is to be given the `latest` tag, but we don't technically have a produced report at that time. Instead, we do the comparison earlier when we're actively using the binary/image to test known-good charts, and verify their reports.

Signed-off-by: Jose R. Gonzalez <jrg@redhat.com>

Related: [HELM-422](https://issues.redhat.com/browse/HELM-422)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /.idea/
 /out/
 **/reports/
+__pycache__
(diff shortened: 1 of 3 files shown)


mmulholla (approved):
LGTM

PR #314, opened 2023-02-28
Bugfix/crds disabled subchart

fenech:
Prevent disabled subcharts containing CRDs from causing tests to fail.

Add a test chart which initially fails, since it contains a **disabled** subchart with a CRD.

By explicitly processing the dependencies (a workaround for [this Helm issue](https://github.com/helm/helm/issues/11852)), the check for CRDs passes.


fenech:
I created this draft PR with a reproducible test case, but I can also open an issue separately if preferred.

To reproduce the failure on the first commit, I built the `chart-verifier` image locally using the Makefile, and ran:

```sh
podman run --rm -it -e KUBECONFIG=/.kube/config \
  -v "${HOME}/.kube":/.kube:z \
  -v "$PWD/internal/chartverifier/checks/":/charts:z \
  "quay.io/redhat-certification/chart-verifier:be2cebd" \
  verify /charts/chart-0.1.0-v3.with-subchart/testchart/
```

Running the same command on the last commit passes the "no CRDs" check:
```
    - check: v1.0/not-contains-crds
      type: Mandatory
      outcome: PASS
      reason: Chart does not contain CRDs
```

Issue #296, opened 2022-12-09
Failed to certify images chart requires kubeVersion: >=1.21 which is incompatible with Kubernetes v1.20.0


imdad000:
we have added kubeVersion: >=1.21 on Chart.YAML
but with chart verifier getting this Error
check: v1.0/images-are-certified
      type: Mandatory
      outcome: FAIL
      reason: 'Failed to certify images : Failed to get images, error running helm template : chart requires kubeVersion: >=1.21 which is incompatible with Kubernetes v1.20.0'

imdad000:
https://stackoverflow.com/questions/74794746/helm-charts-on-ocp-developer-catalog-using-red-hat-repository-issues-with-check

nikeshmi:
@mmulholla Can someone from the RH team look at this issue on priority as its blocking our push to the OpenShift developer catalog

nikeshmi:
Can anyone in the team provide an update on this issue?

mmulholla:
@nikeshmi  This looks an error from helm if you run ```helm template``` on the chart.  Would like to see how kubeVersion is specified in the chart.

mmulholla:
@nikeshmi it should work if you upgrade helm to 3.11

morstad:
chart verifier
```
Digest: sha256:1897b6d0fdc6c8f5dbf6ce2041501f262f3bdeea3469a0c4a7105ce49f3572b0
Status: Image is up to date for quay.io/redhat-certification/chart-verifier:latest
quay.io/redhat-certification/chart-verifier:latest
```

helm version
```
version.BuildInfo{Version:"v3.11.1", GitCommit:"293b50c65d4d56187cd4e2f390f0ada46b4c4737", GitTreeState:"clean", GoVersion:"go1.18.10"}
```

helm template works as expected 
```
helm template https://github.com/IBM/charts/blob/master/repo/ibm-helm/ibm-object-storage-plugin-2.2.10.tgz?raw=true --set license=true
---
# Source: ibm-object-storage-plugin/templates/flex-driver-sa.yaml
# ServiceAccount for ibmcloud-object-storage-drivertrue
apiVersion: v1
kind: ServiceAccount
metadata:
...
```

mmulholla:
I figured out the issue, will have pull request ready soon.

mmulholla:
@morstad I have a PR for a fix #315 

[...]

PR #315, opened 2023-03-01
fix image certify error for kubeVersion >=1.21

mmulholla:
Fix for #296.

The problem was that before calling helm template we pass helm our capabilities which include the kube version supported. Unfortunately we used helms default capabilities which are hard coded to v1.20 and this was the cause of the issue. To fix the problem we now set the kube version in the capabilities on this basis:

1. if an open-shift cluster is available (as would be used for the chart-testing check), we set the capability as the kube version of that cluster. This make sense as the chart kube version should include the version to be tested on. 

2. if an open-shift cluster is not available, set the capability as the highest kube-verision the chart verifier current supports (as described in https://github.com/redhat-certification/chart-verifier/blob/main/internal/tool/kubeOpenShiftVersionMap.yaml). This is just a backup for testing. 

Note: Option 2 would always cover open ended kube versions (>=) but could fail for a specific version or range.

also included a fix to output the cause of a docker build failure if one occurs (as it did for me when working on this fix)


internal/tool/kubectl_test.go
```diff
@@ -50,6 +50,9 @@ var output125 = version.Info{
 	Major: "1",
 	Minor: "25",
 }
+
+var latestVersion = output125
```
komish, line 54: I'd rather do without this declaration here because it is a test file. As the versions change, it would seem we would need to change this variable's value. Is there any other way we can reliably get this value to accomplish this test?

komish (commented):
Overall, LGTM. Just the single comment about a static value in test files.

internal/tool/kubectl_test.go
komish: I'll have to think more on the latter option. For now, unit testing catching a failure to move this value at least allows us to make sure we don't fail to update the test as the version changes. That should suffice for now.

komish (approved):
/lgtm

PR #319, opened 2023-04-03
Extend makefile to add new go developer tooling helpers and downloaders

komish:
Adds Makefile Targets For:

- Installing `gofumpt`, `golangci-lint`, and `gosec`
- Running `gofumpt` instead of `go fmt`. This isn't fully wired up so that GitHub Actions execute them because I believe we should apply them to the codebase before we force them using Actions.


jsm84 (approved):
LGTM

PR #323, opened 2023-04-03
Run gofumpt and golanci-lint against the cmd package

komish:
Running auto-formatters and auto-stylers as a starting point. Doing this package by package to see how they fare against testing.


jsm84 (approved):
/lgtm

PR #324, opened 2023-04-07
Apply gofumpt and golanci-lint to the pkg/report package

komish:
Just applying the styling tools to the report package here. A change was made to a function outside of this package because it calls the function within this package.


jsm84 (approved):
/lgtm

PR #326, opened 2023-04-07
Apply gofumpt and golangci-lint to reportsummary package

komish:
Most of these are style checks for things like `Uri` to `URI`. Few nolints were handed out here.


komish:
This package's tests were looking for exact string matches for errors generated in the source code. Tests were initially failing because the code no longer returned the identical string as expected in tests. The tests have been updated to coincide with the linter's style guideline.

jsm84 (approved):
/lgtm

jsm84 (approved):
/lgtm

PR #328, opened 2023-04-10
Apply gofumpt to checks package

komish:
This package also had no golangci-lint hits, so only gofumpt is being applied here.


jsm84 (approved):
/lgtm

PR #342, opened 2023-04-25
Apply Lint to the remainder of the repo

komish:
This catches anything I missed. After this, we should be able to add `make lint` and `make fmt` to our CI.


jsm84 (approved):
Sorry this took so long but it LGTM.

PR #344, opened 2023-04-26
Enforce style/lint practices on PR

komish:
Update our GitHub actions to use the new make targets for `fmt` and `lint`


jsm84 (approved):
LGTM. Nice and simplified.

PR #345, opened 2023-05-02
Add support for OCP 4.13

vikasmulaje:
Added  kube version and OCP version


komish:
We're unable to add this mapping until a GA release of OCP has landed.

PR #348, opened 2023-05-17
Allow for skipping tls verification with new cluster

komish:
The new cluster is spun up using ACM and has local-only certs. Until we decide to build out a cert pipeline, we'll skip TLS verification for now, which should unblock the CI testing


komish:
This needs to be force-merged because the workflow failing is what's being fixed.

jsm84 (approved):
LGTM

PR #352, opened 2023-06-07
updating dependencies and build processes to latest versions

acornett21:
# Motivation
I noticed inconsistency across projects that were kind of confusing. Also, since this project is now a dependency of `openshift-preflight` the dependencies should be insync so preflight can be built without issues.

# Changes
- Updated the GH actions to use latest actions
- Updated dependencies to use latest
  - Updated go to latest version
- Updated dockerfile to use upstream builder, to get ride of project specific builder image
- Updated the make file, to no longer rely on the `hack/build-*.sh` files
- Updated a code file since a variable changed case sensitivity in a dependency


komish (approved):
/lgtm

Issue #353, opened 2023-06-07
Chart's `value.schema.json` with remote `$ref`s should not be allowed.


tumido:
OCP console seems to be unable to render remote references in JSON schemas within the Helm Catalog. It is a completely valid schema, however, the OCP console can't render it. We should not allow charts with such schemas to enter the OCP Helm catalog.

OCP bug: https://issues.redhat.com/browse/OCPBUGS-14665

I know this is just a workaround, it doesn't catch charts imported by users themselves (via `ProjectHelmChartRepository`), but at least the curated charts should be checked for this.

komish:
Hey @tumido, huge thanks for putting this bug report, linking the context, and throwing together a PR!

After reading through the OCP bug report, I'm unsure that it makes sense to make a chart-verifier change here. The key reason is because remote references are valid schema definition components, as you highlight.

We need and try to strike a balance between checks that enrich the consuming interfaces (e.g. making sure certain annotations exist in the chart definition so the Developer Console has a better UX), and checks that cost chart maintainers extra time/steps and time. If it's valid to include `$refs` in your schema definition, then we wouldn't want to force a certified chart maintainer to remove those from their schema definitions in order to certify, thereby costing them time ensuring this happens before they make submissions.

To that end, I do believe that the Console should not fail in the way that it does. I'm just unsure if this is the right place to fix it, even if it's temporarily until the console workarounds are put in place.

Let me know your thoughts!

komish:
@tumido I believe I see the OCP console bug has been resolved. I'm marking this closed, but please let us know if you believe we should do something else here.

PR #354, opened 2023-06-07
feat: add a check for remote references in values.schema.json

tumido:
Resolves https://github.com/redhat-certification/chart-verifier/issues/353
Related to: https://issues.redhat.com/browse/OCPBUGS-14665, https://github.com/janus-idp/helm-backstage/pull/66

- OpenShift dev console Helm chart catalog doesn't support `$ref`s other than those that start with `#` (local reference), all remote references should flag the chart as failing.
- Adds a `NotContainValuesSchemaRemoteRef` check
- Creates new profile version including this new check
- This check is **Manadatory** for Community as well - I'm not sure how to handle this, but I don't think Community charts should be allowed to break the Dev Console
- The check is based on regex, you may want to change this to traversing the JSON schema - I just though this is much simpler way to check it, if you have concerns I will try spending more time here.
- I'm not sure if I've updated the profiles/default profile/profile tests properly.

Please provide feedback if I'm heading in the right direction.


komish:
Closing this as the associated OCP bug has been resolved.

PR #358, opened 2023-06-20
Re-add removed clone of main during CI tests

komish:
The clone of the `main` branch was removed in the dep cleanup pr: https://github.com/redhat-certification/chart-verifier/pull/352

It was likely removed as a duplicate, but its existence is intentional, so this PR re-adds it at action version 3 (instead of 2), and reconfigured the setup-go action to use main's mod file.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,10 @@ jobs:
         runs-on: ubuntu-latest
 
         steps:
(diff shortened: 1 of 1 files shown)


komish:
FWIW, I expect this to fail on the "setup python for the main branch" because this fixes a problem in the CI itself. To that end, I'll force-merge this once that's confirmed.

PR #359, opened 2023-06-21
Disable CGO on build

komish:
We don't seem to require CGO from what I can see. The golang builder image we use has a newer C toolchain than what's available in RHEL8. If disabling CGO isn't a problem, then this seems like the more compact route to take.

If this doesn't work out in the long term, we can just build from a RHEL8 golang builder image and that should also fix the problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /tmp/src
 
 COPY . .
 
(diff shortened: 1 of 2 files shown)


jsm84 (approved):
lgtm

acornett21 (approved):

jsm84 (approved):
lgtm

PR #363, opened 2023-06-30
Make dev releases prereleases and not drafts

### LINKED ISSUE #355: The dev_release workflow only creates the development 0.1.0 release as a draft
Because the release is not published, the openshift-helm-charts development will throw errors because it can't install the tool.

https://github.com/redhat-certification/chart-verifier/blob/main/.github/workflows/dev_release.yaml

komish:
We randomly run into issues with our dev release, which is used in the certification pipeline's stage, development, and sandbox repositories, because the the release is marked as a draft.

This should that issue, and also specifies that the release should be marked as a prerelease which should address some UX concerns with the GitHub Release UI.

Fixes #355


komish:
This has been rebased off of #364. That needs to be merged first as it contains fixes to CI that are blocking this PR.

jsm84 (approved):

PR #364, opened 2023-06-30
change expected tag after latest tag was unpublished

komish:
It turns out the postgres-10-rhel7 image's[1] latest tag is no longer returned from the Catalog API. I'm not sure if it got unpublished, or what happened, but our test is expecting that image to be there. 

This PR changes the target image to another tag that's currently available (and has been for the last 11 months)

[1]https://catalog.redhat.com/software/containers/rhscl/postgresql-10-rhel7/5aa63541ac3db95f196086f1


komish:
This required changing a bunch of the fixtures that also had references to the same image with the tag latest. For this reason, you'll see other files modified in this PR, but they aren't exactly easy to evaluate. I've just changed their image references to point to the 1-151 tag for postgresql-10-rhel7.

jsm84 (approved):

PR #370, opened 2023-07-10
Add commit ID to version output

mgoerens:
- add a "build" make target, which injects the git commit ID using ldflags
- add the commit ID to the version output
- add the possibility to get the version and commit information in a JSON format

Example output:
```
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ ./chart-verifier version
chart-verifier v1.12.0 <commit: 1c1d51406d0241cbf9683b8dd731eb4bcb706dec>
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ ./chart-verifier version --as-data
{"version":"1.12.0","commit":"1c1d51406d0241cbf9683b8dd731eb4bcb706dec"}[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 IMAGE_BUILDER?=podman
 IMAGE_REPO?=quay.io/redhat-certification
 COMMIT_ID=$(shell git rev-parse --short HEAD)
+COMMIT_ID_LONG=$(shell git rev-parse HEAD)
 
 default: bin
 
@@ -33,7 +34,9 @@ fmt: install.gofumpt
 
 .PHONY: bin
 bin:
-	CGO_ENABLED=0 go build -o ./out/chart-verifier main.go
+	CGO_ENABLED=0 go build \
+		-ldflags "-X 'github.com/redhat-certification/chart-verifier/cmd.CommitIDLong=$(COMMIT_ID_LONG)'" \
+		-o ./out/chart-verifier main.go
 
 .PHONY: lint
 lint: install.golangci-lint
@@ -41,7 +44,9 @@ lint: install.golangci-lint
 
 .PHONY: bin_win
 bin_win:
-	env GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -o .\out\chart-verifier.exe main.go
+	env GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build \
+		-ldflags "-X 'github.com/redhat-certification/chart-verifier/cmd.CommitIDLong=$(COMMIT_ID_LONG)'" \
+		-o .\out\chart-verifier.exe main.go
 
 .PHONY: test
 test:
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -11,8 +12,25 @@ import (
 	apiversion "github.com/redhat-certification/chart-verifier/pkg/chartverifier/version"
 )
 
+// Print version and commit ID as json blob
+var asData bool
+
 func init() {
 	rootCmd.AddCommand(NewVersionCmd())
+}
+
+// CommitIDLong contains the commit ID the binary was build on. It is populated at build time by ldflags.
+// If you're running from a local debugger it will show an empty commit ID.
+var CommitIDLong string = "unknown"
+
+type VersionContext struct {
+	Version string `json:"version"`
+	Commit  string `json:"commit"`
+}
+
+var Version = VersionContext{
+	Version: apiversion.GetVersion(),
+	Commit:  CommitIDLong,
 }
 
 func NewVersionCmd() *cobra.Command {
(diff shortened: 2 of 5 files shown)


komish:
@mgoerens thanks for this. The linter is grumpy ^ https://github.com/redhat-certification/chart-verifier/actions/runs/5508530811/jobs/10039939259?pr=370#step:7:154

mgoerens:
This should be fixed now.

mgoerens:
Fixes #370

Makefile
```diff
@@ -47,6 +48,10 @@ bin_win:
 test:
 	go test -v ./...
 
+.PHONY: build
+build:
+	go build -ldflags "-X 'github.com/redhat-certification/chart-verifier/cmd.CommitIDLong=$(COMMIT_ID_LONG)'"
+
```
komish, line 54: Lets use the existing `bin` target instead of having a separate `build` target. Also add this to `bin_win` for now. 

I'll collapse those targets at some point in the future, but for now, we can duplicate the logic in both places.

cmd/version.go
```diff
@@ -11,23 +12,50 @@ import (
...
+var Version = VersionContext{
+	Version: apiversion.GetVersion(),
+	Commit:  CommitIDLong,
+}
+
 func NewVersionCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "print the chart-verifier version information",
+		Long:  "git commit information for this particular binary build is included at build time and can be accessed by this command",
```
komish, line 40: We can leave the long description out

cmd/version.go
```diff
@@ -11,23 +12,50 @@ import (
...
+	cmd.Flags().BoolVar(&asData, "as-data", false, "output the version and commit ID information in JSON format")
 	return cmd
 }
 
 func runVersion(out io.Writer) error {
-	fmt.Fprintf(out, "v%s\n", apiversion.GetVersion())
+	if asData {
+		marshalledVersion, err := json.Marshal(Version)
+		if err != nil {
+			return err
```
komish, line 54: Instead of returning an error here, what do you think about falling back to the regular way of reporting the version? Maybe with an extra trailing statement saying something like "request for data format failed"?

komish:
@mgoerens Left a couple of comments.

The CI complaint is an interesting way. Effectively what it's implying is that the chart-verifier version output is used in report generation, and since we've changed its output, we're breaking our fixtures. Normally, I'd say we can update our fixtures, however I think for this case, I think we want to make sure we're backwards compatible with the reports because changing that has some hefty implications.

Please let me know if you need any guidance with this. I'd also recommend you are rebased off of main at any given time before submitting your PR, just to rule out any differences between your branch and main when it comes to CI. It can be a little finicky with differences between a PR and main.

cmd/version.go
```diff
@@ -11,23 +12,50 @@ import (
...
+		}
+		fmt.Fprint(out, string(marshalledVersion))
+	} else {
```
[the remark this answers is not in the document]
mgoerens: Ok, I was thinking about this and was balanced. Good that you have a strong opinion on this, that helps making a decision :)

mgoerens:
The code does still make use of pkg/chartverifier/version.GetVersion

```golang
include apiversion "github.com/redhat-certification/chart-verifier/pkg/chartverifier/version"

[...]

var Version = VersionContext{
	Version: apiversion.GetVersion(),
	Commit:  CommitIDLong,
}
```

I ran `make test` prior to submitting the PR and was happy to get a green light. I didn't realize their are also a set of python tests ?

I'll look at option a) and b) in details tomorrow.

komish:
> The code does still make use of pkg/chartverifier/version.GetVersion

Aha, missed this. Thanks for pointing it out!

mgoerens:
All right. Pipeline passes. Though I still don't manage to reproduce it locally, the error seemed to be coming from the lack of newline in the output of `version --as-data`. I can't understand why, but only *some* docker tests were failing due to that, and not always the same were failing depending on which one were run. For instance in [this pipeline](https://github.com/redhat-certification/chart-verifier/actions/runs/5576140326/jobs/10186989815?pr=370) where I commented out the first docker test + all podman and tarball tests, only the 5th and 6th test pass (out of seven). If I run all tests only the first docker test passes, and following docker tests fail (all podman and tarball tests pass regardless). Go figure ... Besides, I cannot reproduce this behavior on a fresh debian instance.

So I added a newline and now all tests pass.

Full list of changes (from commit message):

    - use ldflags to injects the git commit ID at build time
    - add the commit ID to the version output
    - add the possibility to get the version and commit information in a
      JSON format
    - adapt the tests to make use of the JSON output
    - adapt Dockerfile to make use of the make bin target
    - fix path to chart used to test the built docker image in buildandtest
      script (moved as part of !271) and make use of the JSON output

Reg. the last point: the "build-and-test" script is, as the name suggests, comprised of two parts: first it builds the container image, then it tests the image using `verify` on a known chart. The script is called with the `--build-only` flag in the pipeline i.e. the image is not tested using this script.

There are multiple issues with the "test" part which makes this script not usable as-is (prior to this PR):
- the chart used for testing has been moved to another location (easy fix)
- the verify command uses a `-l` flag which isn't supported anymore (easy fix)
- the verify command requires access to a Kubernetes cluster. Currently there is no way to provide Kubernetes credentials to the script (fixable, with some effort)

But before putting further effort into fixing the "test" part of this script, it appears to me like we don't need this script to be able to test a docker image anymore. The pipeline takes care of that in a next step (actually running some extensive tests). AFAICT this script is never called without the `--build-only` flag in our testing / release process (and anyway it would not have worked, even prior to this PR).

My suggestion is to *not* fix this part and to consider dropping the "test" part of the `buildandtest` script altogether in another PR.

mgoerens:
Opened https://github.com/redhat-certification/chart-verifier/issues/378 for the buildandtest issue

cmd/version.go
komish: Fair point there. LGTM

komish (approved):
/lgtm

Thanks @mgoerens!

PR #382, opened 2023-07-20
Move release logic to a new workflow

### LINKED ISSUE #380: Commit ID in version output doesn't match a commit on the main branch
```
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ ~/Downloads/chart-verifier version
chart-verifier v1.12.2 <commit: 4b15f462cb269cccfa5ddc81ec4fb7a3bb14d978>
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ git log 4b15f462cb269cccfa5ddc81ec4fb7a3bb14d978
fatal: bad object 4b15f462cb269cccfa5ddc81ec4fb7a3bb14d978
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ git show 4b15f462cb269cccfa5ddc81ec4fb7a3bb14d978
fatal: bad object 4b15f462cb269cccfa5ddc81ec4fb7a3bb14d978
[mgoerens@mgoerens-thinkpadp1gen3 chart-verifier]$ git show HEAD -q
commit aacc2c152001c90b5e8336ed6440bb20f7704fb0 (HEAD -> main, tag: 1.12.2, origin/main, origin/HEAD)
Author: Jose R. Gonzalez <jose@flutes.dev>
Date:   Tue Jul 18 15:50:19 2023 -0500

    Release 1.12.2 (#379)
```

mgoerens:
- remove steps creating the GitHub release from "build.yaml" workflow.
- add steps to tag the main branch with the version tag after an
  automerge in the "build.yaml" workflow.
- create a new GitHub Action for handling GitHub release on new tags on
  main.
- modify the python script handling the release to be able to get the
  release body directly.

fixes #380


scripts/src/release/releasebody.py
```diff
@@ -0,0 +1,25 @@
+import sys
+
+sys.path.append('./scripts/src/')
+from release import releasechecker
+from utils import utils
+
+def make_release_body(version, image_name, release_info):
```
komish, line 7: Since we're making modifications here (even if we're just moving this), could I have you add docstrings?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  build-and-push:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
```
komish, line 17: I believe we need to setup Go here.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
+name: create release on new tags
+
+on:
+  push:
+    # Publish semver tags as releases.
+    tags: [ '*.*.*' ]
```
komish, line 6: We'll want to make sure this works how we want. I think you mentioned you'd test this in a dev repo, I'd like to see that when you have it standing.

.github/workflows/release.yaml
komish: To clarify, I believe you need it before you `make bin` in the next step.

.github/workflows/build.yaml
```diff
@@ -200,21 +200,23 @@ jobs:
...
               if: ${{ steps.check_version_updated.outputs.updated == 'true'}}
               run: |
-                ve1/bin/check-auto-merge --api-url=${{ github.event.pull_request._links.self.href }}
+                git fetch
+                export ORIGIN_MASTER_SHA=$(git rev-parse origin/master)
+                echo "origin_master_sha=$ORIGIN_MASTER_SHA" >> $GITHUB_OUTPUT
 
-            - name: Create the the release
-              id: create_release
+            - name: Tag master
```
komish, line 211: NIT: I'd probably rename this to `create release tag`.
It's also worth noting that it isn't clear what should/would happen if that tag existed already. I believe that also makes this task not re-runnable once we get to this point?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          make bin
+          tar -zcvf chart-verifier.tgz -C out/ chart-verifier
+          export TARBALL_PATH=$(realpath chart-verifier.tgz)
+          echo "tarball_path=$TARBALL_PATH"
+
+      - name: Print tag to GITHUB_OUTPUT
+        id: get_tag
+        run: |
+          echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
```
komish, line 29: I don't remember exactly what shows up in GitHub Action logs, but if this won't get displayed to stdout, maybe we should use `tee -a` instead of `>>` so that we also see it in stdout of this task.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Set up Python scripts
+        run: |
+          # set up python requirements and scripts on PR branch
+          python3 -m venv ve1
+          cd scripts && ../ve1/bin/pip3 install -r requirements.txt && cd ..
+          cd scripts && ../ve1/bin/python3 setup.py install && cd ..
+
+      - name: Generate release body
+        id: release_body
+        run: ve1/bin/print-release-body
```
komish, line 45: This is a little unclear - are we just printing the release body or are we storing it. Should be sending this to `GITHUB_OUTPUT`? If we are doing this in-script, I'd love to see that be a flag so that we can run this on our machine without the expectation of running within the GHA environment. E.g. `--to-github-output`

.github/workflows/release.yaml
mgoerens: Good point. Surprisingly in my tests, the `make bin` step ran successfully without having to setup go beforehand. I'll still add this step.

.github/workflows/release.yaml
mgoerens: I've changed it to something a little bit more robust (`[0-9]+.[0-9]+.[0-9]+`). Note that we don't prefix our version with `v`.

I've tested using a random string and the pipeline is not triggered.

.github/workflows/release.yaml
mgoerens: That's a good idea, I changed it to use "tee -a"

.github/workflows/build.yaml
mgoerens: That is correct, if the tag already exists then this step will fail.

I think we can live with this "limitation" for the following reasons:
- IMO this step should be the last step of this workflow. Currently it is followed by the "retagging" to `latest` of the quay.io container image. But ideally those steps should be moved to the "release.yaml" workflow so that all assets (tarball and container image) are created there. To be discussed of course, but if we agree this is something I would address in a separate PR to keep this one a little bit lighter.
- It's unlikely that we want to re-release a particular version. But if we need to, we can manually delete and recreate the tag. This will recreate the GitHub release (old one will become a "Draft" that needs to be manually deleted) and the tarball. Once the container image build is moved to the release workflow, they will also be recreated by this action.

.github/workflows/release.yaml
mgoerens: I've changed it to print to stdout and use `tee -a` to get it to GITHUB_OUTPUT too.

mgoerens:
@komish This is ready for another round of review.

I've tested in this repo: https://github.com/mgoerens/my-chart-verifier. It's an almost one to one copy of the chart-verifier repo, in particular in terms of GitHub workflows. If there is any scenario you want to check we can do so there. I've tested in particular:
- Creating a standard PR (not a release): Run the tests but doesn't automerge / tag/
- Creating a release pipeline: Run the tests + automerge + tag
- Upon tagging of the main branch: create the release + tarball
- Delete tag: Release becomes Draft release
- Create tag manually: triggers creation of release + tarball

A few notes:
* As mentioned, this allows the recreation of a GitHub release and its associated assets (currently only tarball, later also container image) by manually deleting and recreating the tag (`git -d 1.12.2 && git push --delete origin 1.12.2 && git tag 1.12.2 && git push --tags`)
* In that case however, the tests are not run. Also the "old" release is not completely deleted but continues to live as a Draft. It needs to be manually cleaned up.
* In the build.yaml workflow, the steps which tags the main branch is run using a Personal Access Token (currently set to `PTA_MGOERENS`, but that secret doesn't exist in this repo). This is because by default, a GitHub Workflow cannot trigger another GitHub Workflow. See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

mgoerens:
Here's an example release PR in `mgoerens/my-chart-verifier`: https://github.com/mgoerens/my-chart-verifier/pull/3

The associated Workflow runs:
- in the PR (note this is red because I haven't configured the appropriate Build triggers on Quay.io for this image, but the failure happens after the PR is merged and master is tagged): https://github.com/mgoerens/my-chart-verifier/actions/runs/5626007182/job/15246830087
- on master after automatic tagging: https://github.com/mgoerens/my-chart-verifier/actions/runs/5626464345

And the release it created: https://github.com/mgoerens/my-chart-verifier/releases/tag/1.13.0

komish:
@mgoerens it likely is a conflict if the same chart happened to be installed at the same time. Multiple pipeline runs can occur because things happen in namespaced contexts. I'll take a look at the underlying cluster here shortly.

.github/workflows/build.yaml
```diff
@@ -200,21 +205,23 @@ jobs:
...
               if: ${{ steps.check_version_updated.outputs.updated == 'true'}}
-              uses: softprops/action-gh-release@v1
+              uses: mathieudutour/github-tag-action@v6.1
               with:
-                tag_name: ${{ steps.check_version_in_PR.outputs.PR_version }}
-                body: ${{ steps.check_version_in_PR.outputs.PR_release_body }}
-                files: ${{ steps.check_version_in_PR.outputs.PR_tarball_name }}
-              env:
-                GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                github_token: ${{ secrets.PAT_MGOERENS }}
```
komish, line 221:
```suggestion
                github_token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,79 @@
+name: create release on new tags
+
+# This workflow is triggered by new version tags (e.g. "1.12.2") on the main branch.
+#
+# Tagging should be automatic after each merged PR release (see the build.yaml workflow). It can
+# also be manually triggered if needed. Note that the tests are not run by this workflow, so
+# release manually at your own risk. If manually tagging the main branch, the tag must match the
+# version set in ./pkg/chartverifier/version/version_info.json.
```
komish, line 8: Just so I'm clear, when you say this can be manually triggered, you mean by creating a tag?
In other words, there's no `workflow_dispatch` event that this triggers on, and that's normally what I see when a workflow can be manually triggered.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod
+
+      - name: Print tag to GITHUB_OUTPUT
+        id: get_tag
+        run: |
+          echo "release_version=${GITHUB_REF#refs/*/}" | tee -a $GITHUB_OUTPUT
```
komish, line 34: Maybe it's being too cautious, but should we also consider printing out the `git rev-parse HEAD` output?

komish:
Looking good overall. I want to do a few more tests and look over your repo before I mark this approved.
The only other thing I can think of that might be problematic is the dev release. Just want to make sure that's doing what I think it should even after these changes.

.github/workflows/release.yaml
mgoerens: That's correct, it can be triggered ~manually~ by manually creating a tag on the main branch with the right format (x.x.x without leading v).

It's not intended to be triggered manually via the GitHub GUI / CLI.

I'll reword the comment.

.github/workflows/release.yaml
mgoerens: I see you already resolved this conversation ? (intended ?)

I don't think it's needed to print the commit ID, unless you have a strong opinion otherwise

.github/workflows/build.yaml
mgoerens: Sorry I'm a bit too verbose sometimes, I mentioned in a previous comment:

> In the build.yaml workflow, the steps which tags the main branch is run using a Personal Access Token (currently set to PTA_MGOERENS, but that secret doesn't exist in this repo). This is because by default, a GitHub Workflow cannot trigger another GitHub Workflow. See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

So I believe we must use a Personal Access Token. We'll have to create the secret in this repo before merging (before we cut a new release). I'm happy to rename it to something else than `PAT_MGOERENS` though.

komish (approved):
/lgtm

komish:
Looks great. We'll need to review the next "release" that includes changes to chart-verifier itself to make sure everything looks as we expect.

PR #385, opened 2023-07-25
Check images for an empty image string

### LINKED ISSUE #347: Charts should fail cleanly when an deployment resource's image value is set to an empty string
E.g. `image: ' '` being found in the deployment spec. This seemed to cause the ImagesAreCertified check to hang indefinitely, which was causing a failure in our CI.

The submission that helped us find this bug is here: https://github.com/openshift-helm-charts/charts/pull/815/files#diff-005e794a0836903a45fa4321bb10ea105f445d383ec230c67d426132058f13e8R42

If the image field is empty, we should not hang. I'm not certain if I prefer that we just ignore the value and rely on installation-based tests to assert that this should fail, or if we should fail the certification check altogether. Either way, we should not hang.

For what it's worth, there were two tests that appeared to be causing a hang in chart-verifier's execution. If this is being addressed and a hang is not observed, just making sure the above criteria is met should suffice.

OchiengEd:
Update the `getImagesFromContent` function to return all images, including empty ones.

In addition, when evaluating images received in certifyImages() function, skip evaluating registries for any empty images

These changes also have a performance improvement. Unlike the previous implementation of the function that evaluated each line of the generated helm templates for images.

```
chart-verifier/ (empty-image✗) $ go test -v  -run=^$ -bench=^BenchmarkGetImagesFromContent  github.com/redhat-certification/chart-verifier/internal/chartverifier/checks
goos: darwin
goarch: amd64
pkg: github.com/redhat-certification/chart-verifier/internal/chartverifier/checks
cpu: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
BenchmarkGetImagesFromContent
BenchmarkGetImagesFromContent-8      	    2780	    429554 ns/op
BenchmarkGetImagesFromContentAlt
BenchmarkGetImagesFromContentAlt-8   	   18494	     64014 ns/op
PASS
ok  	github.com/redhat-certification/chart-verifier/internal/chartverifier/checks	4.940s
chart-verifier/ (empty-image✗) $
```

Closes #347

### DIFF
--- a/internal/chartverifier/checks/helm.go
+++ b/internal/chartverifier/checks/helm.go
@@ -25,6 +25,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"helm.sh/helm/v3/pkg/chartutil"
 
@@ -223,12 +224,21 @@ func getImageReferences(chartURI string, vals map[string]interface{}, kubeVersio
 // getImagesFromContent evaluates generated templates from
 // helm and extracts images which are returned in a slice
 func getImagesFromContent(content string) []string {
-	var images []string
-	re := regexp.MustCompile(`\s+image\:\s+(?P<image>.*)`)
+	re := regexp.MustCompile(`\s+image\:\s+(?P<image>.*)\n`)
 	matches := re.FindAllStringSubmatch(content, -1)
+	imageMap := make(map[string]struct{})
 	for _, match := range matches {
-		images = append(images, match[re.SubexpIndex("image")])
+		image := strings.TrimSpace(match[re.SubexpIndex("image")])
+		image = strings.Trim(image, "\"")
+		image = strings.Trim(image, "'")
+		imageMap[image] = struct{}{}
 	}
+
+	var images []string
+	for k := range imageMap {
+		images = append(images, k)
+	}
+
 	return images
 }
 
--- a/internal/chartverifier/checks/helm.go
+++ b/internal/chartverifier/checks/helm.go
@@ -218,13 +218,16 @@ func getImageReferences(chartURI string, vals map[string]interface{}, kubeVersio
 		return nil, err
 	}
 
-	return getImagesFromContent(txt), nil
+	return getImagesFromContent(txt)
 }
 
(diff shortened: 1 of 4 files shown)


internal/chartverifier/checks/helm.go
```diff
@@ -221,52 +217,19 @@ func getImageReferences(chartURI string, vals map[string]interface{}, kubeVersio
...
-		for isPrefix && err == nil {
-			var partLine []byte
-			partLine, isPrefix, err = reader.ReadLine()
-			nextLine = append(nextLine, partLine...)
-		}
+// getImagesFromContent evaluates generated templates from
+// helm and extracts images which are returned in a slice
+func getImagesFromContent(content string) []string {
+	var images []string
+	re := regexp.MustCompile(`\s+image\:\s+(?P<image>.*)`)
```
komish, line 227: MustCompile makes this a panic if it doesn't, right? If so, can we store this string as a constant and ensure MustCompile works in our unit tests to catch a panic?

internal/chartverifier/checks/helm_test.go
```diff
@@ -180,7 +180,7 @@ func TestLongLineTemplate(t *testing.T) {
 	content, err := os.ReadFile("templates/test-template.yaml")
 	require.NoError(t, err)
 
-	images, err := getImagesFromContent(string(content))
+	images, err := getImagesFromContent(string(content)), nil
```
komish, line 183: Injecting `nil` here looks odd to me. This is because our return values changed in the signature?

Also, should we extend our test coverage to include an empty image value check?

komish (changes requested):
Just a few questions ^

internal/chartverifier/checks/helm_test.go
OchiengEd: Correct. The function signature of the `getImagesFromContent()` function changed. So in an effort to avoid changing a ton of things, left that as a `nil` for now.

internal/chartverifier/checks/helm.go
OchiengEd: This error has been handled in the latest commit.

PR #386, opened 2023-07-26
Move container images creation to release workflow

### LINKED ISSUE #383: Build container images in the release workflow instead of using Build triggers
This ensures all release assets are built in the same workflow. This also allows to retag the main branch in order to recreate all those assets at once if needed.

mgoerens:
This makes it possible to retrigger the entire release process, including all assets creation, by retagging the main branch.

Starting with this commit, build triggers on the
quay.io/redhat-certification/chart-verifier repo are no longer needed.

close #383

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,7 +18,8 @@ name: create release on new tags
 # Finally, as mentioned above, create a new tag (e.g. `git tag 1.12.2 && git push --tags`) to
 # recreate the release.
 #
-# This workflow builds the tarball, creates the Github release and attaches the tarball to it.
+# This workflow builds all release assets (the tarball and the container images), creates the
+# Github release and attaches the tarball to it.
 
 on:
   push:
@@ -88,3 +89,21 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Build container images
+        id: build_container_images
+        run: |
+            # Build podman images locally
+            make build-image IMAGE_TAG=${{ steps.get_tag.outputs.release_version }}
+            make build-image IMAGE_TAG=latest
+
+      - name: Push to quay.io
+        id: push_to_quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: chart-verifier
+          tags: |
+            ${{ steps.get_tag.outputs.release_version }}
+            latest
+          registry: quay.io/redhat-certification
+          username: ${{ secrets.QUAY_BOT_USERNAME }}
+          password: ${{ secrets.QUAY_BOT_TOKEN }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,7 +3,7 @@ name: Build, Test, Automerge and Tag
 # This workflow runs on all PRs that are targetting the main branch.
 #
 # It runs the test suite. If the PR is a release PR, it automerges and tags the main branch with
-# the corresonding new version. Finally, it retags the container image with latest.
+# the corresonding new version.
 
 on:
     pull_request_target:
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -88,3 +89,25 @@ jobs:
...
+
+      - name: Push to quay.io
+        id: push_to_quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: chart-verifier
+          tags: |
+            ${{ steps.get_tag.outputs.release_version }}
+            latest
+            main
```
komish, line 110: This syntax looks strange. The doc does seem to imply you can specify multiple tags separated by whitespace, so I'm sure this works, but wow it looks odd! (Feel free to resolve this, as there's no action I'm really looking to have done here)

komish: Actually, I lied - I'm pretty sure we don't want to tag main here. Maybe latest. This is a release workflow that triggers on tags. Main and often will contain commits that are beyond a given release.

.github/workflows/release.yaml
```diff
@@ -88,3 +89,25 @@ jobs:
...
 
+      - name: Build container images
+        id: build_container_images
+        run: |
+            # Build podman images locally
+            image_tag=${{ steps.get_tag.outputs.release_version }}
+            podman build \
+              -t quay.io/redhat-certification/chart-verifier:$image_tag \
+              -t quay.io/redhat-certification/chart-verifier:latest \
+              -t quay.io/redhat-certification/chart-verifier:main .
```
komish, line 100: Same thing here re: main (see other comment for details)

.github/workflows/release.yaml
```diff
@@ -88,3 +89,25 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Build container images
+        id: build_container_images
+        run: |
+            # Build podman images locally
+            image_tag=${{ steps.get_tag.outputs.release_version }}
+            podman build \
```
komish, line 97: Do we want this to exist as a make target?

komish (changes requested):
Couple of things to consider mentioned in comments. Otherwise, looks pretty good. Once merged, I'll go disconnect the Quay auto-build.

Oh! You might want to remove the build status from README.md.

.github/workflows/release.yaml
mgoerens: Sorry about the syntax :upside_down_face: I agree it's not super pretty, but I'd say I've seen worse.

You're right, I'll remove building and pushing of `main`. I'm wondering if we should keep the Build Trigger or create a new Github Workflow to build the `main` image.

To clarify, we have 4 types of images:
- `x.y.z` (i.e. version tags): built as part of the release process (currently using Quay Build Triggers, replaced in this PR by a step in "release.yaml" worfklow)
- `latest`: follows the latest `x.y.z`. Currently retagged at the end of the "build.yaml" on release, moved in this PR to "release.yaml"
- `0.1.0` (i.e. dev release): built on each push to the main branch, currently in "dev_release.yaml" workflow. I think we should leave this untouched.
-  `main`: built for each push on the main branch via a Quay Build Trigger.

So, AFAICT the `0.1.0` and the `main` tag are always in sync. How do we want to build `main` ? We have following options:
1. Leave the current Build Trigger untouched.
2. Create a new GitHub workflow specifically for building the `main` image (triggered on each push to the main branch)
3. Reuse the "dev_release.yaml" workflow as it is triggered for each push to the main branch already, and integrate building the main image there.

Not a big fan of 3. personally, as I prefer to separate concerns. 1. means that we keep a build trigger active, and I was hoping to get rid of all Build Triggers here (not that I dislike Build Triggers, but I'd prefer having all the builds happening in the same place i.e. GitHub Actions). That leaves us with 2., even though I'd prefer not to create yet another workflow, I believe that's reasonable.

komish: Option 2 or 3 is the right way forward. We don't have to make them the same workflow, we can just make the existing workflow for 0.1.0 reusable. Easy enough!

komish: Clarity: Dealing with the main tag doesn't need to happen in this PR. We can do it at a later time. The build trigger won't go away until this lands anyway.

mgoerens: I added a workflow for tagging the main branch in this PR.

.github/workflows/release.yaml
mgoerens: Created a make target

komish (approved):
/lgtm

Build triggers have been disabled in Quay.

PR #393, opened 2023-07-31
Remove testing using docker

### LINKED ISSUE #378: Remove docker-related tests and docs
~There are multiple issues with the "test" part which makes this script not usable as-is:~
~- the chart used for testing has been moved to another location (easy fix)~
~- the verify command uses a -l flag which isn't supported anymore (easy fix)~
~- the verify command requires access to a Kubernetes cluster. Currently there is no way to provide Kubernetes credentials to the script (fixable, with some effort)~

~But before putting further effort into fixing the "test" part of this script, it appears to me like we don't need this script to be able to test a docker image anymore. The pipeline takes care of that in a next step (actually running some extensive tests). AFAICT this script is never called without the --build-only flag in our testing / release process and anyway would not work.~

~Hence suggesting to remove the "test" part of this script.~

See below

mgoerens:
The container image test is already covered by the podman tests.

- Remove testing the container image using docker.
- Update the documentation accordingly
- Remove the "build-and-test" python script, previously used to build the docker image.
- In places where the podman image is built, use the "build-image" make target.

close #378


komish:
@mgoerens this needs a rebase. The actions/cache@v4 tag didn't exist, so we replaced it with the v3 tag which does and merged that to resolve the setup issue.

mgoerens:
@komish FYI the pipeline fails as the workflows from the `main` branch are being run, and the PR removes some files essentials for the docker test to run.

komish:
@mgoerens understood. Makes sense. We can test the behavior here in a fork without issue.

mgoerens:
Tested in mgoerens/my-chart-verifier:
- I applied all changes made by this branch in the main branch of mgoerens/my-chart-verifier
- (Used sed to rename to the correct repo)
- created a fake release PR https://github.com/mgoerens/my-chart-verifier/pull/9
- docker tests are removed, pipeline is green.

komish (approved):
Successful run replicated in my repo as well.

https://github.com/komish/chart-verifier/actions/runs/5742028147

/lgtm

PR #395, opened 2023-07-31
Update gitignore file to skip system files

OchiengEd:
Update the `.gitignore` to ignore system and other golang artifacts
* Generate `coverage.out`  when the test target, (i.e. `make test`), is executed
* Other minor updates / optimizations


komish (approved):
/lgtm

PR #400, opened 2023-08-03
Remove unecessary steps in dev_release workflow

mgoerens:
Python steps are not used anymore in this workflow, hence those setup steps can now be removed.


komish (approved):
/lgtm

PR #401, opened 2023-08-03
Add flake8 step in the build pipeline

mgoerens:


komish:
Does this step _fail_ when the code doesn't match style guidelines? Just want to confirm as applying them here won't do anything, and we want the step to fail so the contributor can make their fixes and repush.

mgoerens:
I now installed flake8 in a dedicated venv.

komish (approved):
/lgtm

PR #402, opened 2023-08-03
Expire commit container images after 1 week

### LINKED ISSUE #397: Add expiration date to "commit tags" on quay
Since #386 we push one new tag to Quay.io per commit on the `main` branch. This creates a lot of new tags, that we probably won't really need down the road.

Some numbers of https://quay.io/repository/redhat-certification/chart-verifier?tab=tags:
* In total, there are currently 33 tags
* In the last 6 days (since #386 was merged), 7 new tags have been created (one for each commit on `main`).

image

At this rhythm, we're going to have hundreds of "commit tags" in just a few months.

This is a suggestion to add an expiration to those images. See https://docs.projectquay.io/use_quay.html#tag-expiration

Maybe 1 month is enough ?

mgoerens:
Prior to podman 4.6.0, there is a bug that would cause the expiration tag to be added to the "main" container image as well in the build.yaml workflow. See the buildah bug report:
https://github.com/containers/buildah/issues/4632

As a workaround, the --no-cache flag is used to build container images with the "build-image" make target.

This workaround, can be removed when we use a runner image that has podman>=4.6.0. Currently, the ubuntu-latest runner image we use has 3.4.4.

close #397

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ IMAGE_REPO?=quay.io/redhat-certification
 COMMIT_ID=$(shell git rev-parse --short HEAD)
 COMMIT_ID_LONG=$(shell git rev-parse HEAD)
 IMAGE_TAG=$(COMMIT_ID)
+QUAY_EXPIRE_AFTER="never"
 
 default: bin
 
@@ -57,13 +58,18 @@ test:
 # If IMAGE_TAG is not provided, use the COMMIT_ID
 .PHONY: build-image
 build-image:
-	$(IMAGE_BUILDER) build -t $(IMAGE_REPO)/chart-verifier:$(IMAGE_TAG) .
+	# TODO: Adding --no-cache option as a workaround to https://github.com/containers/buildah/issues/4632
+	# This can be removed as soon as we can ensure that the ubuntu-latest runner image uses podman=>4.6.0
+	$(IMAGE_BUILDER) build \
+		--no-cache \
+		--label quay.expires-after=$(QUAY_EXPIRE_AFTER) \
+		-t $(IMAGE_REPO)/chart-verifier:$(IMAGE_TAG) .
 
 # Push the container image. Usage: make push-image IMAGE_TAG=my_tag
 # If IMAGE_TAG is not provided, use the COMMIT_ID
 .PHONY: push-image
 push-image:
-	$(IMAGE_BUILDER) push $(IMAGE_REPO)/chart-verifier:$(IMAGE_TAG) .
+	$(IMAGE_BUILDER) push $(IMAGE_REPO)/chart-verifier:$(IMAGE_TAG)
 
 .PHONY: gosec
 gosec: install.gosec
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -25,8 +25,9 @@ jobs:
       - name: Build container images
         id: build_container_images
         run: |
+            podman version
             # Build podman images locally
-            make build-image IMAGE_TAG=${{ steps.get_commit_id.outputs.commit_id }}
+            make build-image IMAGE_TAG=${{ steps.get_commit_id.outputs.commit_id }} QUAY_EXPIRE_AFTER=1w
             make build-image IMAGE_TAG=main
 
       - name: Push to quay.io


komish:
Well that's a bit weird! Do you think we should use the buildah image in our CI[1], or stick with this approach which allows us to keep the core build logic in our Makefile?

[1]https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/.github/workflows/build-main.yml#L31-L43

mgoerens:
I think this is a fine workaround. I'm aware of the buildah step that is used in preflight, but I personally like better to build all our images using the make target ("keep the core build logic in our Makefile" as you well put it).

> --no-cache: do not use existing cached images for the container build. Build from the start with a new set of cached layers.

So there is a performance hit for using this flag, which IMO is acceptable. If we think this it too annoying (e.g. we run `make build-image` in our daily workflows locally, and have to wait a considerable amount of time), we can also make this flag optional, something like:

Makefile:
```
[...]
BUILD_FLAGS=""

[...]
build_image:
    podman build $BUILD_FLAGS -t ....
```

and in build.yaml:
```
make build_image BUILD_FLAGS="--no-cache" ...
```

But I'm not sure this is even necessary... Building images is mostly done is the pipeline, and it's okay if it takes one more minute IMO.

mgoerens:
FYI I just amended the commit to improve the readability of the make target with line breaks.

Makefile
```diff
@@ -57,7 +58,12 @@ test:
 # If IMAGE_TAG is not provided, use the COMMIT_ID
 .PHONY: build-image
 build-image:
-	$(IMAGE_BUILDER) build -t $(IMAGE_REPO)/chart-verifier:$(IMAGE_TAG) .
+	# TODO: Adding --no-cache option as a workaround to https://github.com/containers/buildah/issues/4632
+	# This can be removed as soon as we can ensure that the ubuntu-latest runner image uses podman=>4.6.0
+	$(IMAGE_BUILDER) build \
+		--no-cache \
+		--label quay.expires-after=$(QUAY_EXPIRE_AFTER) \
```
komish, line 65: Let's try one more thing to see if we can avoid having to skip the cache on this older version of Podman.

Instead of labeling this here, maybe lets label it in the dockerfile as we do here: https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/Dockerfile#L38

And then pass the expiration as a build argument: 
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/Dockerfile#L7

I'm thinking this might sidestep the bug because the LABEL command would produce a different layer due to having a different input.

komish (changes requested):
Overall LGTM, but just one more thing to try before we commit to using `--no-cache` to work around this bug.

Makefile
mgoerens: Tested locally, unfortunately that doesn't work, even with the latest podman version, which is surprising:

```
$ podman build --build-arg quay_expiration=1w -t test_expire_1w .
[...]
Successfully tagged localhost/test_expire_1w:latest

$ podman build --build-arg quay_expiration=never -t test_no_expire .
[...]
Successfully tagged localhost/test_no_expire:latest
Successfully tagged localhost/test_expire_1w:latest

$ podman inspect test_no_expire:latest | grep quay.expires-after
                    "quay.expires-after": "1w",
               "quay.expires-after": "1w",
                    "created_by": "/bin/sh -c #(nop) LABEL quay.expires-after=${quay_expiration}",
```

mgoerens: Also just tested with using `buildah build` rather than podman, but I get the same behavior.

komish: Thanks for spending time trying out some options here. A couple of relatively unstructured thoughts here after doing some testing:

I don't particularly like removing an image locally to fix this problem. Overall, it's fine, but it feels a bit unnatural for what we're trying to achieve here (simply pushing this image with various tags). To that end, I don't think we need to do that here, and would rather we compromise in some other way to accomplish our goal here. Let's revert back on this one.

I think I also strongly dislike using Podman 3.x in our GHA environment. Not for any particular reason, just that I really like the state of Podman 4.x, and want to make sure we're getting the latest fixes. In addition, there'd be environmental variance between our GHA and our local workstations, which for the most part should be on Podman 4.x. This is mostly a tangent, and there's no action to take here.

With that said, I'm not sure there's much we can do aside from building out our own runners (which just feels like overkill). I'm uncertain if the Podman- and Buildah-related image building actions get us newer version of these tools, but that doesn't matter here.

So here's where I think I land on this:

One of my goals is/was to get the `:main` tag and the `:$commitsha` tags to basically be identical, but just have one expire (`$commitsha`) and the other move periodically (as `main` updates). When I say Identical, I mostly mean their image digests are the same, which is easier to do if we just retag (or `skopeo copy`) the same image into the multiple places. 

The problem ends up being the way we're telling Quay to expire the image, which is through the use of a label that we inject at image _build_ time.

I don't think we technically _need_ to do this, and we could just hit the Quay API to do this, which would potentially solve the problem. That is, 

1. Build Image based on main
2. skopeo copy that image to the corresponding commit hash
3. Hit the Quay API to tell the commit hash tag to expire after N days.

An example of this is available in the [Clair](https://github.com/quay/clair/blob/main/.github/actions/set-image-expiration/action.yml) repo.

Now... with all that said, I totally think that's too much effort for something that's going to expire / move. So. I I'll drop that goal. With that in mind, I think we can just simplify this to get it merged. 

We could:

1) Build it how you did originally, relying on `--no-cache` to get past the bug in Podman (simplest)
2) Look into a reusable action [1] that separates the build of the various tags (which means all of the image digests will differ, but as I mentioned, that's fine - this goes away after our expiration anyway)

I'd say choose whichever path you like, and we'll go with it. As long as our release tags are persistent, I think what we do with main and commit tags doesn't have to absolutely "perfect", whatever that may mean. And I don't want you to have to continue trying things out here, when all of these options work perfectly fine. 

[1] This is what I tested, but I'm not sure it solves any particular problems other than ensuring that at a given tag is pushed without conflict on other image build processes. It would need to be reconfigured to use Make targets if we're interested in this route.
https://github.com/komish/test-action-podman-build/blob/main/.github/workflows/workflow-call-image-create.yml
https://github.com/komish/test-action-podman-build/actions/runs/5791519331

mgoerens: I suggest to keep the `--no-cache` flag as a temporary measure for simplicity, with a nice TODO explaining the issue it solves.

The next Ubuntu LTS will be Ubuntu 24.04, to be released in April 2024. I rely on it having a podman version>=4.6.0 i.e. that contains the bug fix. The `ubuntu-latest` runner will then use 24.04. At this moment, the `--no-cache` flag can be dropped.

komish: Let's do it.

PR #404, opened 2023-08-09
Build dev release container image in main

### LINKED ISSUE #391: Reconfigure dev release workflow to use podman
https://github.com/redhat-certification/chart-verifier/blob/main/.github/workflows/dev_release.yaml

In https://github.com/redhat-certification/chart-verifier/pull/386 we reconfigured our image building workflow to leverage podman. The dev release was configured to use the scripts that leveraged the docker API in order to match the main workflow. Since that's now changed, the dev release can probably be updated accordingly.

As a secondary item, the dev release `0.1.0` and the `main` image tags are built on the same triggers, but are built at different times. It may make sense to work these two things together such that they're the same underlying image.

mgoerens:
close #391

Example of what images are produced:

image

### DIFF
--- a/.github/workflows/dev_release.yaml
+++ b/.github/workflows/dev_release.yaml
@@ -1,7 +1,7 @@
 name: Create Dev Release
 # Delete the previous development release and recreate it
-# with code based off of the main branch. Also updates the
-# DEV_RELEASE tag for the image to match.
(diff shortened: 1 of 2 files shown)


dcurran90 (approved):
lgtm

dcurran90 (approved):

komish (approved):
/lgtm

PR #405, opened 2023-08-10
Adding support for daemonSet and statefulSet resources in charts

### LINKED ISSUE #286: Add support for testing daemon and statefulsets
Hello,

Please add support for testing daemon and statefulsets on test releases.

On [this ](https://github.com/redhat-certification/chart-verifier/blob/ac429dd185bc67df81e0e924435c1e33e2323c67/internal/chartverifier/checks/charttesting.go#L223) line, chart-verifier waits only for Deployments, but if test runs against daemon and statefulsets, they are not guaranteed to be deployed when helm tests start.

Thank you!

dcurran90:
Update waitForDeployments to waitForWorkloads and add the ability to wait for DaemonSets and StatefulSets.
Add tests for this new feature and increase code coverage

Closes #286

### DIFF
--- a/internal/chartverifier/checks/charttesting.go
+++ b/internal/chartverifier/checks/charttesting.go
@@ -235,7 +235,7 @@ func testRelease(
 	release, namespace, releaseSelector string,
 	cleanupHelmTests bool,
 ) error {
-	if err := kubectl.WaitForDeployments(ctx, namespace, releaseSelector); err != nil {
+	if err := kubectl.WaitForWorkloadResources(ctx, namespace, releaseSelector); err != nil {
 		return err
 	}
 	if err := helm.Test(ctx, namespace, release); err != nil {
--- a/internal/tool/kubectl.go
+++ b/internal/tool/kubectl.go
@@ -100,20 +100,22 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
 func (k Kubectl) WaitForWorkloadResources(context context.Context, namespace string, selector string) error {
 	deadline, _ := context.Deadline()
 	unavailableWorkloadResources := []workloadNotReady{{Name: "none", Unavailable: 1}}
-
 	getWorkloadResourceError := ""
 
+	// Loop until timeout reached or all requested pods are available
 	utils.LogInfo(fmt.Sprintf("Start wait for workloads resources. --timeout time left: %s ", time.Until(deadline).String()))
 	for deadline.After(time.Now()) && len(unavailableWorkloadResources) > 0 {
 		unavailableWorkloadResources = []workloadNotReady{}
 
-		deployments, errDeployment := listDeployments(k, context, namespace, selector)
+		deployments, errDeployments := listDeployments(k, context, namespace, selector)
 		daemonSets, errDaemonSets := listDaemonSets(k, context, namespace, selector)
 		statefulSets, errStatefulSets := listStatefulSets(k, context, namespace, selector)
 
-		if errDeployment != nil {
+		// Else/If block will handle errors from API requests for each
+		// resource type or inspect the resources that are successfully returned
+		if errDeployments != nil {
 			unavailableWorkloadResources = []workloadNotReady{{Name: "none", ResourceType: "Deployment", Unavailable: 1}}
-			getWorkloadResourceError = fmt.Sprintf("error getting deployments from namespace %s : %v", namespace, errDeployment)
+			getWorkloadResourceError = fmt.Sprintf("error getting deployments from namespace %s : %v", namespace, errDeployments)
 			utils.LogWarning(getWorkloadResourceError)
 			time.Sleep(time.Second)
 		} else if errDaemonSets != nil {
@@ -128,9 +130,8 @@ func (k Kubectl) WaitForWorkloadResources(context context.Context, namespace str
 			time.Sleep(time.Second)
 		} else {
 			getWorkloadResourceError = ""
+			// Check the number of unavailable replicas for each workload type
 			for _, deployment := range deployments {
-				// Just after rollout, pods from the previous deployment revision may still be in a
-				// terminating state.
 				if deployment.Status.UnavailableReplicas > 0 {
 					unavailableWorkloadResources = append(unavailableWorkloadResources, workloadNotReady{Name: deployment.Name, ResourceType: "Deployment", Unavailable: deployment.Status.UnavailableReplicas})
 				}
@@ -141,13 +142,15 @@ func (k Kubectl) WaitForWorkloadResources(context context.Context, namespace str
 				}
 			}
 			for _, statefulSet := range statefulSets {
-				// TODO: Double check that Replicas is requested and Available is currently running
+				// StatefulSet doesn't report unavailable replicas so it is calculated here
 				unavailableReplicas := statefulSet.Status.Replicas - statefulSet.Status.AvailableReplicas
 				if unavailableReplicas > 0 {
 					unavailableWorkloadResources = append(unavailableWorkloadResources, workloadNotReady{Name: statefulSet.Name, ResourceType: "StatefulSet", Unavailable: unavailableReplicas})
 				}
 			}
 
+			// If any pods are unavailable report it and sleep until the next loop
+			// If everythign is available exit the loop
 			if len(unavailableWorkloadResources) > 0 {
 				utils.LogInfo(fmt.Sprintf("Wait for %d workload resources:", len(unavailableWorkloadResources)))
 				for _, unavailableWorkloadResource := range unavailableWorkloadResources {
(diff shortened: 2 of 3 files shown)


internal/tool/kubectl_test.go
```diff
@@ -193,19 +336,76 @@ func deploymentTestListGood(k Kubectl, context context.Context, namespace string
 	return DeploymentList1, nil
 }
 
+func daemonSetTestListGood(k Kubectl, context context.Context, namespace string, selector string) ([]v1.DaemonSet, error) {
```
mgoerens, line 339: Could we add some comments to this function and the next ones ?

internal/tool/kubectl.go
```diff
@@ -94,49 +97,76 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
 	return kubectl, nil
 }
 
-func (k Kubectl) WaitForDeployments(context context.Context, namespace string, selector string) error {
+func (k Kubectl) WaitForWorkloadResources(context context.Context, namespace string, selector string) error {
```
mgoerens, line 100: I think it would be a good occasion to comment this function

mgoerens (changes requested):
Looks good overall, thanks ! I've added some small suggestions.

internal/tool/kubectl.go
```diff
@@ -94,49 +97,76 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
...
+			unavailableWorkloadResources = []workloadNotReady{{Name: "none", ResourceType: "StatefulSet", Unavailable: 1}}
+			getWorkloadResourceError = fmt.Sprintf("error getting stateful sets from namespace %s : %v", namespace, errStatefulSets)
+			utils.LogWarning(getWorkloadResourceError)
 			time.Sleep(time.Second)
 		} else {
-			getDeploymentsError = ""
+			getWorkloadResourceError = ""
 			for _, deployment := range deployments {
 				// Just after rollout, pods from the previous deployment revision may still be in a
 				// terminating state.
```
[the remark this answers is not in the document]
dcurran90: good call. I'll double check all the comments and make them more relevant.

internal/tool/kubectl.go
dcurran90: comments added to describe the various sections of the function

komish: @dcurran90 I think @mgoerens was looking for a top-level comment E.g. 
```suggestion
// WaitForWorkloadResources loops until ...
func (k Kubectl) WaitForWorkloadResources(context context.Context, namespace string, selector string) error {
```

Don't know why that diff looks so awful, but either way, refer to the commenting structure defined here under "Funcs" https://go.dev/doc/comment

internal/chartverifier/checks/charttesting.go
```diff
@@ -235,7 +235,7 @@ func testRelease(
 	release, namespace, releaseSelector string,
 	cleanupHelmTests bool,
 ) error {
-	if err := kubectl.WaitForDeployments(ctx, namespace, releaseSelector); err != nil {
+	if err := kubectl.WaitForWorkloadResources(ctx, namespace, releaseSelector); err != nil {
```
komish, line 238: https://go.dev/doc/comment

internal/tool/kubectl.go
```diff
@@ -94,49 +97,80 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
...
+		} else if errDaemonSets != nil {
+			unavailableWorkloadResources = []workloadNotReady{{Name: "none", ResourceType: "DaemonSet", Unavailable: 1}}
+			getWorkloadResourceError = fmt.Sprintf("error getting daemon sets from namespace %s : %v", namespace, errDaemonSets)
+			utils.LogWarning(getWorkloadResourceError)
+			time.Sleep(time.Second)
+		} else if errStatefulSets != nil {
+			unavailableWorkloadResources = []workloadNotReady{{Name: "none", ResourceType: "StatefulSet", Unavailable: 1}}
+			getWorkloadResourceError = fmt.Sprintf("error getting stateful sets from namespace %s : %v", namespace, errStatefulSets)
+			utils.LogWarning(getWorkloadResourceError)
 			time.Sleep(time.Second)
```
komish, line 130: A few things for this block.

1) This feels like we should be dropping the `else if` component. Running through each `if` block separately should work here.

2) There's a good deal of repetition in what's being executed, indicating we could probably just refactor this to remove it and, in turn, be more concise.

internal/tool/kubectl.go
```diff
@@ -94,49 +97,80 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
...
+			// If everythign is available exit the loop
+			if len(unavailableWorkloadResources) > 0 {
+				utils.LogInfo(fmt.Sprintf("Wait for %d workload resources:", len(unavailableWorkloadResources)))
+				for _, unavailableWorkloadResource := range unavailableWorkloadResources {
+					utils.LogInfo(fmt.Sprintf("    - %s %s with %d unavailable pods", unavailableWorkloadResource.ResourceType, unavailableWorkloadResource.Name, unavailableWorkloadResource.Unavailable))
 				}
 				time.Sleep(time.Second)
 			} else {
-				utils.LogInfo(fmt.Sprintf("Finish wait for deployments, --timeout time left %s", time.Until(deadline).String()))
+				utils.LogInfo(fmt.Sprintf("Finish wait for workload resources, --timeout time left %s", time.Until(deadline).String()))
```
komish, line 161: 1) The structure here feels like we can abstract out the parsing of these into separate functions and call them.
2) Perhaps there's an opportunity to use goroutines here to let the parsing of statefulsets, deployments, and daemonsets work independently of each other? WDYT. I think it could do away with the sleeps I'm seeing.

internal/tool/kubectl.go
```diff
@@ -94,49 +97,80 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
...
+	getWorkloadResourceError := ""
+
+	// Loop until timeout reached or all requested pods are available
+	utils.LogInfo(fmt.Sprintf("Start wait for workloads resources. --timeout time left: %s ", time.Until(deadline).String()))
+	for deadline.After(time.Now()) && len(unavailableWorkloadResources) > 0 {
+		unavailableWorkloadResources = []workloadNotReady{}
+
+		deployments, errDeployments := listDeployments(k, context, namespace, selector)
+		daemonSets, errDaemonSets := listDaemonSets(k, context, namespace, selector)
+		statefulSets, errStatefulSets := listStatefulSets(k, context, namespace, selector)
```
komish, line 112: So I'm understanding that we check for these three resources regardless of whether the deployed chart uses them? The idea being that charts that don't use statefulsets/daemonsets would just auto-pass the "readiness" check because none of those workloads exist. Is that a correct understanding?

That's fine, I just want to make sure. Fundamentally, this check seems to be flawed because charts can deploy _all sorts_ of resources. But at the end of the day, this is a better check than nothing.

komish (changes requested):
@dcurran90 Thanks for your work on this.

Few comments left. I think we can be a little more concise here, and doing so would preserve the readability of these bits. Let me know if I can help with anything.

internal/tool/kubectl.go
dcurran90: yes, charts that don't try do deploy certain resources will just auto-pass since we're not expecting them to be there.

As far as checking for other resources I could add in jobs at least because it seems like the only one we're missing from the list of default workload resources (https://kubernetes.io/docs/concepts/workloads/) but didn't know if i should since it wasn't requested and didn't want to go overboard. Happy to make another PR for that though if it's needed

internal/tool/kubectl.go
```diff
@@ -94,49 +97,80 @@ func NewKubectl(kubeConfig clientcmd.ClientConfig) (*Kubectl, error) {
...
+
+		// Else/If block will handle errors from API requests for each
+		// resource type or inspect the resources that are successfully returned
+		if errDeployments != nil {
+			unavailableWorkloadResources = []workloadNotReady{{Name: "none", ResourceType: "Deployment", Unavailable: 1}}
+			getWorkloadResourceError = fmt.Sprintf("error getting deployments from namespace %s : %v", namespace, errDeployments)
+			utils.LogWarning(getWorkloadResourceError)
+			time.Sleep(time.Second)
```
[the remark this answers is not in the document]
dcurran90: This was actually left over from what the function previously did. 
I believe the idea is to avoid excessive calls and rate limiting since it'll immediately hit the API again after these failures. Happy to remove it if this is a non-issue but trying to find the default rate limit info first

internal/tool/kubectl.go
dcurran90: updated comments.

komish:
@dcurran90 just a heads up that tests are failing here, and should be reproducible to `make test`

komish (approved):
/lgtm

@mgoerens when you're happy with this, go ahead and merge away.

mgoerens (approved):

PR #408, opened 2023-09-29
Use external library to get the OCP versions range

mgoerens:
The translation of range of Kubernetes version to range of OCP versions was previsouly inconsistent accross chart-verifier and the certification pipeline. The code that performs this translation has now been externalized to the github.com/opdev/get-ocp-range repository.

See https://github.com/openshift-helm-charts/charts/issues/1032


internal/chartverifier/checks/checks_test.go
```diff
@@ -549,51 +549,6 @@ func TestRequiredAnnotationsPresent(t *testing.T) {
 	}
 }
 
-func TestSemVers(t *testing.T) {
```
komish, line 552: Any particular reason this test needs to change? Has this test been moved to the replacing library?

mgoerens: These are the unit tests for the `getOCPRange` function, which is removed from this repo with this commit.

The Go library implements unit tests too, see https://github.com/opdev/getocprange/blob/main/get-ocp-range_test.go

komish (approved):
/lgtm

PR #410, opened 2023-10-11
remove output directory from file list for digest

### LINKED ISSUE #316: Chart digest generation is impacted if you run chart-verifier from within the same directory as the chart
## What Happened

I noticed that chart-verifier was producing a report with `.metadata.tool.digests.chart` changing with every execution, despite pointing to the same location with no changes. 

When executing it, I was in the chart directory running `chart-verifier verify .`.

This was producing the `chartverifier/` directory which contained log files. With each iteration, a new log file would be produced in this directory.

These log files are getting picked up when the chart location is slurped up, and as a result, the sorted list of files contained in the chart was changing every execution, causing my calculated digest to change (ultimately failing certification as well).

## Detail

The code snippet of concern is here:

https://github.com/redhat-certification/chart-verifier/blob/main/internal/chartverifier/reportBuilder.go#L196. The `rawFiles` parameter contains all files in that directory so those are all taken into account when calculating the hash.

The resolution would probably either need to decipher what's necessary and what isn't.

## Version

v1.10.0

## Workaround

It would seem that running chart-verifier in a different directory should produce an accurate hash. I'm still testing this now.

dcurran90:
Added a filter function to remove output files created by chart-verifier from the list of files used to calculate the sha256 digest

Closes #316

### DIFF
--- a/internal/chartverifier/reportBuilder.go
+++ b/internal/chartverifier/reportBuilder.go
@@ -169,6 +169,20 @@ type fileSorter struct {
 	by    func(p1, p2 *helmchart.File) bool // Closure used in the Less method.
 }
 
+// If chart-verifier is run from within the helm chart directory
+// the output will be sent to the OutputDirectory and affect the digest.
+// This removes any verifier output files from the calculated digest
+func filterOutputDirectory(files []*helmchart.File) []*helmchart.File {
+	n := 0
+	for _, file := range files {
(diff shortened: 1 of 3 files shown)


komish:
Thanks @dcurran90 we'll consider this for the next release

komish (approved):
Working how I expect, thanks @dcurran90 

/lgtm

PR #414, opened 2023-10-31
Bump Golang to v1.21

komish:
Bumping golang, and explicitly bumping the x/net package to 0.16.0.


acornett21:
If you update the latest dot release of k8's dependencies (v0.28.3) you'll get these updates for 'free', though I'm not sure we want to do that here.

PR #421, opened 2023-12-18
Removing unused variable setOverridesFlag

### LINKED ISSUE #322: Identify the source of the unused linting warning for setOverridesFlag
In cmd/verify.go https://github.com/redhat-certification/chart-verifier/blob/main/cmd/verify.go#L47, the `setOverridesFlag` triggers golangci-lint because it's unused. 

We should verify that it's unused, try to find places where its use is intended but disconnected, and decide to either wire it in or remove it. Worth checking the commit where it was introduced, and any commits thereafter to see how it has been manipulated.

dcurran90:
Removing unused variable. [This](https://github.com/redhat-certification/chart-verifier/commit/3f00873be5e34fa1f5749b2eada90d3784fa57ba) commit switched the relevant variable and didn't remove it's initialization.

Closes #322

### DIFF
--- a/cmd/verify.go
+++ b/cmd/verify.go
@@ -51,13 +51,6 @@ var (
 	disabledChecksFlag []string
 	// outputFormatFlag contains the output format the user has specified: default, yaml or json.
 	outputFormatFlag string
-	// TODO(komish): The description of this would imply that it's important. We generally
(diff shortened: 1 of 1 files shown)


komish (approved):
/lgtm

PR #423, opened 2024-02-06
Use get-ocp-range to translate Kube to OCP ranges

### LINKED ISSUE #418: Remove the Kube-to-OCP version mapping file from this repo, instead relying on functionality from opdev/getocprange
We rely on the opdev/getocprange library for calls that translate kube version strings to OCP version strings (e.g. for checking version ranges in Chart.yaml), but this repository still has a [mapping file](https://github.com/redhat-certification/chart-verifier/blob/ceb5d8efbd53ba86e6ce59596a97fbbb0215b3fd/internal/tool/kubeOpenShiftVersionMap.yaml#L4) that needs to be updated with each release.

Instead, adjust functionality that uses this in this source code such that it relies fully on the getocprange library.

mgoerens:
This work had been started as part of commit f08a985e53d90671c6c99d88149390ef3a1c3bfc

This commit removes the mapping of Kubernetes versions to OCP versions and calls get-ocp-range instead.

closes #418


internal/chartverifier/checks/helm.go
```diff
@@ -194,7 +193,7 @@ func getImageReferences(chartURI string, vals map[string]interface{}, kubeVersio
 	capabilities := chartutil.DefaultCapabilities
 
 	if kubeVersionString == "" {
-		kubeVersionString = tool.GetLatestKubeVersion()
+		kubeVersionString = "1.27" // This is *not* a viable solution, to be discussed in PR comments
 	}
```
mgoerens, line 197: See also how this function is called: https://github.com/redhat-certification/chart-verifier/blob/9e4c6ccea2748570d5028926f8743d68a5d7817f/internal/chartverifier/checks/checks.go#L465-L479

In my understanding, there is an attempt (with `tool.GetClientConfig`) to determine the Kubernetes version (that this chart supports ??) based on Environment variable that might be set. I somehow ended [here](https://github.com/helm/helm/blob/v3.14.0/pkg/cli/environment.go#L93) in my investigations.

In case that failed (for instance because the environment variables are not set), then there is a fallback here, that defaults to the latest known Kubernetes version (i.e. the latest set in `internal/tool/kubeOpenShiftVersionMap.yaml`).

I'm wondering:
1. what exactly does this Kubernetes version represents. Is this supposed to be the supported version for this chart ?
2. if defaulting to the latest Kubernetes version really makes sense here.
3. if the fallback serves a real-world scenario or is just to accommodate the tests, when the Helm env vars are not set.

I'll continue pulling those threads, but in the meantime I wanted to post my thoughts here.

komish: Effectively, this kubeVersion value seems to align with the server that's being used for testing, and if the server is not available, then it defaults to the newest version.

The image reference resolution code effectively emulates `helm template`, and seems to be emulating the design of the upstream `action` library: https://pkg.go.dev/helm.sh/helm/v3@v3.14.0/pkg/action.

As I read it, the summary is that the RenderManifests function that ultimately underpins `getImageReferences` takes in the user's helm-relevant input values (such as values.yaml), and then uses a template rendering strategy to derive manifests with their values.yaml-provided images rendered out. Otherwise, images would be templated in the manifests being scanned, and they would all fail (for being invalid). It also makes sense for cases where a partner is rendering one value for upstream Kubernetes, and another values for OpenShift.

As far as *why* a kubeclient is necessary, I'd have to dig deeper. Happy to do that if you'd like. I stopped at this point because I think the end goal is that a cluster being unavailable should not prevent the templating from taking place. I'm presuming that no actual calls to the cluster are made, but the [upstream actions](https://github.com/redhat-certification/chart-verifier/blob/main/internal/helm/actions/template.go#L20) being called ask for an actions.Configuration, which requires a client. As such, the latest version is being subbed in to produce a client, regardless of whether calls to a given cluster are made - and only checks that are related to image reference resolution.

Hope that helps, let me know if I can expand.

mgoerens: Thanks. That helped.

After giving it a fresh look at it I found don't need to set a kubeVersion at all, hence we can remove the problematic logic from the code.

This `kubeVersion` variable is in fact not needed to instantiate a new Kubernetes Client. That is done [here](https://github.com/redhat-certification/chart-verifier/blob/694dfdea0c284f35ae32e827ab9a9a9d9f31cf85/internal/chartverifier/checks/helm.go#L195) using a fake client.

It turns out that the `kubeVersion` that was set only serves an informative purpose. It is fed into the `chartutil.Capabilities` struct. By using the [`chartutil.DefaultCapabilities`](https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L41) we are already provided with a default `kubeVersion`. I don't believe we need to set our own.

komish (approved):
/lgtm

PR #430, opened 2024-04-01
Add NOTES.txt check

### LINKED ISSUE #428: Add check for the existence of  NOTES.txt template in a given chart
In helm chart  that contains `NOTES.txt` template, it rendered and printed to the screen after helm install/upgrade command completed.

It's very recommended to have a NOTES.txt template in any chart, because in this way, we can supply for the user many interesting details and information about the deployment that helm install command just deployed, and how to use it, customize it , and even more.

obviously implementing this check is very  simple and straightforward.

More details of helm `NOTES.txt` template:

https://helm.sh/docs/chart_template_guide/notes_files/

trgeiger:
Checks for NOTES.txt in the chart's templates directory. Includes tests.

Closes #428 

Right now it's just another mandatory check since we don't have any logic for optional checks. I'm not sure how everyone feels about optional vs. mandatory.


komish:
@trgeiger thanks for this - just two things

- NOTES.txt is intended to convey information to the end-user, but some charts may not have information to convey. If we're going to add this check, it probably needs to be optional across all profiles.
- Any references to "Notes" should specify "NOTES.txt", so that our expectations are clear.

internal/chartverifier/checks/checks.go
```diff
@@ -111,6 +113,23 @@ func HasReadme(opts *CheckOptions) (Result, error) {
...
 
+func HasNotes(opts *CheckOptions) (Result, error) {
+	c, _, err := LoadChartFromURI(opts)
+	if err != nil {
+		return Result{}, err
+	}
+
+	r := NewResult(false, NotesDoesNotExist)
+	for _, f := range c.Templates {
+		fmt.Println(f.Name)
```
komish, line 124: Please remove this line

komish (approved):
/lgtm

PR #435, opened 2024-04-24
Case sensitive management for input options keys

### LINKED ISSUE #434: Input options keys should be case sensitive
It looks like the keys provided as input options for the `verify` are managed in a case insensitive way (always lower case), hence preventing them to work as expected when key contains some upper case characters.

For example, if `verify` is executed providing `-S replicaCount=2`, then `replicaCount` becomes `replicacount`.
I've figured this out because I have an Helm chart with `"additionalProperties": false` in the schema and when verifying it, I got:
```
results:
    - check: v1.0/helm-lint
      type: Mandatory
      outcome: FAIL
      reason: |+
        Helm lint has failed:  [INFO] Chart.yaml: icon is recommended
        [ERROR] values.yaml: - (root): Additional property replicacount is not allowed

        [ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
        mychart:
        - (root): Additional property replicacount is not allowed
```

mrizzi:
Fixes #434


komish (approved):
Looks reasonable to me, thank you @mrizzi!

/lgtm

PR #441, opened 2024-06-06
add starting dependabot configuration

komish:
Dropping in a starting dependabot configuration. Note that the scripts directory ("pip" package-ecosystem) is not included as I'm currently evaluating a switch to go-based testing. If I decide not to go that route, I'll add in the pip ecosystem management to this configuration in a future PR.


mgoerens (approved):
LGTM

PR #447, opened 2024-06-07
Update integration test dependencies

komish:
This PR updates the python environment used for integration testing for chart-verifier to more closely match the helm certification repositories. I decided not to pursue rewriting integration tests in Go for the moment given that the scripts that are in project are also used in GHA. In effect, if/when I decide to replace those, I'll move GHA first. I didn't want to change all of this at once (this PR already contains plenty of changes).

Notable:

**scripts/requirements.txt** contains many version updates. This is a one-off large update, but dependabot is being integrated to manage updates more frequently and in smaller doses over time.

**.github/actions/setup-python/**, merged https://github.com/redhat-certification/chart-verifier/pull/448, is being called throughout our workflows. It's intending to simplify how many times we have to specify the python version we use through actions. Mostly a PoC, but works identically to what we had before. We can use pyproject.toml in the future for this, but if we migrated the current value from setup.cfg to pyproject, we'd have a `>=` instead of a pinned version - so for now, I'll pin the version and we can re-evaluate if we want to use the pyproject value to pass into setup-python.

**pyproject.toml** is being added, and related configurations (e.g. setup.cfg) are updated to support its use. Eventually this will be followed with a python version bump.

**.github/dependabot.yml** is being updated to enable dependabot updates for the pip ecosystem.

**chart_good.feature / chart_test.py** have both been updated to account for new changes in both pytest and pytest-bdd. Notably, Feature examples are no longer supported. We got "matrix" style testing that way, so for now, I've just replicated this by adding an extra Example column.


jsm84 (approved):
lgtm

PR #455, opened 2024-07-10
Inline the workflow that generate sbom into the general release tasks

komish:
The github-actions bot is what creates our release, and this bot cannot trigger other workflows. For that reason, our SBOM generation did not trigger because it waits for release events, but it's not being sent because of the previous.

This PR inlines the SBOM generation and attaches it during the release creation.

If SBOM generation fails, we continue as we still want the release to complete.


jsm84 (approved):
LGTM

PR #466, opened 2024-08-14
Fix bug in values merging when ci directory exists

### LINKED ISSUE #465: I can't override my Chart values
I'm trying to verify my chart (https://github.com/SonarSource/helm-chart-sonarqube), but I can't override the default values.

I'm trying to apply [this](https://github.com/SonarSource/helm-chart-sonarqube/pull/526/files#diff-0b1f46ebc6c31e7761ac1eeb0a4fa55d6fe935bfdffa28b6ec8f3a7c4f1fe244) values.yaml:

```yaml
jwtSecret: "dZ0EB0KxnF++nr5+4vfTCaun/eWbv6gOoXodiAMqcFo="

image:
  image:
  pullSecrets:
    - name: pullsecret
  repository: "sonarsource/sonarqube"
  tag: "10.7.0-community"
```

I tried both commands without success:

```
$ docker run --net host -e KUBECONFIG=/.kube/config -v "${HOME}/.kube/":/.kube/ -v "$(pwd)/charts":/charts:z "quay.io/redhat-certification/chart-verifier" -F /charts/sonarqube/ci/openshift-verifier-values.yaml -n sonarqube verify /charts/sonarqube

$ docker run --net host -e KUBECONFIG=/.kube/config -v "${HOME}/.kube/":/.kube/ -v "$(pwd)/charts":/charts:z "quay.io/redhat-certification/chart-verifier" -S image.tag=10.7.0-community (...) -n sonarqube verify /charts/sonarqube
```

In both cases, it will use the [default](https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/values.yaml#L33) value for the tag: `10.6.0-community`. This also applies to any other value. Adding `--debug` doesn't show any information.

What am I missing here?

komish:
We were incorrectly assigning values contents into a transient variable where we intended to capture its value in variable assigned in one of the upper scopes. As a result, if a helm chart happened to have a ci directory (which has significance in chart-testing), then we'd end up with no values at all because we'd serialize an empty struct into a temporary values file and use that for testing.

This PR fixes that and also migrates a mergo function that's deprecated to its replacement implementation.

Fixes #465

### DIFF
--- a/internal/chartverifier/checks/charttesting.go
+++ b/internal/chartverifier/checks/charttesting.go
@@ -371,12 +371,13 @@ func newTempValuesFileWithOverrides(filename string, valuesOverrides map[string]
 
 	if filename != "" {
 		// in the case a filename is provided, read its contents and merge any available values override.
-		obj, err := readObjectFromYamlFile(filename)
+		var err error
+		obj, err = readObjectFromYamlFile(filename)
 		if err != nil {
 			return "", nil, fmt.Errorf("reading values file: %w", err)
 		}
 
-		err = mergo.MergeWithOverwrite(&obj, valuesOverrides)
+		err = mergo.Merge(&obj, valuesOverrides, mergo.WithOverride)
 		if err != nil {
 			return "", nil, fmt.Errorf("merging extra values: %w", err)
 		}


jsm84 (approved):
/lgtm

PR #468, opened 2024-08-15
Sign release images

komish:
This PR will run `Cosign` in keyless mode to sign our tagged images (which are associated with the release process). Note that this is not the latest cosign Y-stream version, but is the same version we use in other places and know to work how we expect.

This PR adds an additional GitHub Secret `secrets.IMAGE_REGISTRY` and replaces hard-coded references to the same value. The secret is already configured for the repository. I've also moved the image name to an environment variable in-file. The combination of these changes makes it so you can more easily test this task in a fork (by setting your own secrets).

In addition, this PR will change our strategy for managing the latest tag. The previous implementation would simply call `podman build` twice and push the results. They would end up with different digests. Instead, we call it once, and tag `latest` once the versioned image is built. That way, when pushed, they will be linked.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,11 +26,19 @@ on:
     # Publish semver tags as releases.
     tags: '[0-9]+.[0-9]+.[0-9]+'
 
+env:
+  IMAGE_NAME: chart-verifier
+
 jobs:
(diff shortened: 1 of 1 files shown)


jsm84 (approved):
/lgtm

PR #483, opened 2025-01-15
Add installed oc bin to the path

komish:


komish:
This won't pass CI because of the problem it's solving, but this change doesn't touch the binary build. Merging, further CI testing will happen in one of the many dependency PRs that are open.

PR #515, opened 2025-07-22
Update getocprange dependency to support Go 1.24

caxu-rh:
Now that this codebase is using Go 1.24, we can also pick up getocprange updates, which include (recursive) dependency updates and vulnerability fixes (https://github.com/opdev/getocprange/pull/10).


komish (approved):
/lgtm

PR #522, opened 2025-08-11
workflows: improve security posture

caxu-rh:
- Pin actions used to a Git SHA-1 hash instead of a tag
- Don't persist credentials when using actions/checkout, except where needed (I did not find any places where it is needed)
- Use `gh` CLI to delete/recreate dev release instead of using third-party actions
- Use `gh` CLI to create tagged releases instead of using third-party actions


.github/workflows/release.yaml
```diff
@@ -114,17 +117,18 @@ jobs:
...
-            ${{ steps.generate_sbom_filename.outputs.sbom_filename }}
+      - name: Create tag and release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ steps.get_tag.outputs.release_version }}
+          RELEASE_NOTES: ${{ steps.release_body.outputs.release_body }}
+          TARBALL_PATH: ${{ steps.build_bin.outputs.tarball_path }}
+          SBOM_FILENAME: ${{ steps.generate_sbom_filename.outputs.sbom_filename }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run:
```
komish, line 127:
```suggestion
        run: |
```

Think we should add the `|` multiline indicator here.

I don't think it's strictly necessary because we only have one command in scope, but because it's written with escaped line breaks, I'd say let's go ahead and add it for consistency.

.github/workflows/dev_release.yaml
```diff
@@ -47,30 +49,19 @@ jobs:
...
-              # We think this is because of a race condition.
-              # Force a wait after the previous release was deleted.
-            - name: Sleep and to buffer release recreation
-              id: sleep-after-delete-previous
-              run: sleep 8
+              env:
+                GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                TARBALL_FULL_NAME: ${{ steps.create-tarfile.outputs.tarball_full_name }}
+              run: |
+                gh release delete ${DEV_RELEASE} --cleanup-tag --yes
```
komish, line 58:
```suggestion
                gh release delete "${DEV_RELEASE}" --cleanup-tag --yes
```

komish (approved):
/lgtm

PR #588, opened 2026-04-24
Ensure proper directory prefix for github workflows

komish:
The previous value would result in the dependabot config being unable to find the workflows.


mgoerens (approved):
LGTM

Just for my information, why did you remove the `groups` ? Is it because we'd rather update actions individually for better testing ? We'll have more PRs as a result (which is fine)

komish:
@mgoerens it's fair, I'm not sure how it'll play out (will we get actions/checkout updated across all instances? Just one in a PR? the latter is not great) but the goal was to make sure we change smaller bits because the workflows changes aren't tested until the subsequent PR (IIRC).

PR #589, opened 2026-04-28
ensure we append to GITHUB_OUTPUT and require workflow parameters

komish:
The functional test workflow would only output two of the three required outputs when determining if a PR was valid to test, causing issues in pulling down the content to be tested.

### DIFF
--- a/.github/workflows/functional-tests.yaml
+++ b/.github/workflows/functional-tests.yaml
@@ -16,8 +16,7 @@ on:
           (e.g. github.event.pull_request.number, etc.)
       checkout-repository:
         type: string
-        required: false
-        default: ""
+        required: true
         description: |
           repository flag to actions/checkout.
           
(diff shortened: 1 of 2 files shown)


komish:
Maintainers, note that this is not expected to pass functional tests because it fixes a bug in the functional tests.

caxu-rh:
> Maintainers, note that this is not expected to pass functional tests because it fixes a bug in the functional tests.

Is there a change in the functional tests also needed which would turn this green, or would we just be planning to merge as-is?

jsm84 (approved):
Very concise and straightforward. LGTM

komish:
@caxu-rh The theory is that this change in the workflow is what makes the test that's failing go green. The problem is that the installed version of chart verifier (in the workflow run) is old, and doesn't recognize the new cluster version that's used in CI. The fix is this patch. To me, it's a chicken and egg - these tests aren't green because this patch hasn't merged.

The other open PRs at this moment will be rebased and should pass following this merge. They suffer the same problem.

PR #607, opened 2026-05-11
fix(verifier): return digest error when public key digest fails

caxu-rh:
Previously GetPublicKeyDigest errors were dropped and the outer verify error (which is always nil at this point) was returned instead.


komish (approved):
/lgtm

PR #608, opened 2026-05-11
refactor(chartverifier): simplify GetPackageDigest and document helpers

caxu-rh:
Extract openChartPackageReader for clearer scheme handling and a single defer rc.Close() on success. Propagate os.Open errors instead of ignoring them. Add godoc for GetPackageDigest and openChartPackageReader.

### DIFF
--- a/internal/chartverifier/reportBuilder.go
+++ b/internal/chartverifier/reportBuilder.go
@@ -223,30 +223,40 @@ func GenerateSha(rawFiles []*helmchart.File) string {
 	return fmt.Sprintf("sha256:%x", chartSha.Sum(nil))
 }
 
+// openChartPackage opens the chart package byte stream at u as an [io.ReadCloser].
+func openChartPackage(u *url.URL) (io.ReadCloser, error) {
+	switch u.Scheme {
+	case "http", "https":
+		resp, err := http.Get(u.String())
+		if err != nil {
+			return nil, err
+		}
+		return resp.Body, nil
+	case "file", "":
+		if !strings.HasSuffix(u.Path, ".tgz") {
+			return nil, nil
+		}
+		return os.Open(u.Path)
+	default:
+		return nil, fmt.Errorf("scheme %q not supported", u.Scheme)
+	}
+}
+
+// GetPackageDigest returns a hex-encoded SHA256 hash of the chart package bytes at uri.
 func GetPackageDigest(uri string) string {
-	url, err := url.Parse(uri)
+	u, err := url.Parse(uri)
 	if err != nil {
 		return ""
 	}
-	var chartReader io.Reader
-	switch url.Scheme {
-	case "http", "https":
-		var chartGetResponse *http.Response
-		chartGetResponse, err = http.Get(url.String())
-		if err == nil {
-			chartReader = chartGetResponse.Body
-		}
-	case "file", "":
-		if strings.HasSuffix(url.Path, ".tgz") {
-			chartReader, _ = os.Open(url.Path)
-		}
-	default:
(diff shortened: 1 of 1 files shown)


internal/chartverifier/reportBuilder.go
```diff
@@ -223,30 +223,45 @@ func GenerateSha(rawFiles []*helmchart.File) string {
...
-		return ""
-	}
-	var chartReader io.Reader
-	switch url.Scheme {
+// openChartPackageReader opens the chart package byte stream at u for digesting.
+// On success, it returns an [io.ReadCloser]: for http or https, the response body from [http.Get];
+// for file or an empty scheme, the file from [os.Open] when the URL path ends with ".tgz".
+// It returns (nil, nil) for file or an empty scheme when the path does not end with ".tgz".
+// It returns (nil, err) if the scheme is unsupported, or if [http.Get] or [os.Open] fails.
+func openChartPackageReader(u *url.URL) (io.ReadCloser, error) {
```
komish: nit: I would drop `Reader` from this function name.

internal/chartverifier/reportBuilder.go
```diff
@@ -223,30 +223,45 @@ func GenerateSha(rawFiles []*helmchart.File) string {
...
-	if err != nil {
-		return ""
-	}
-	var chartReader io.Reader
-	switch url.Scheme {
+// openChartPackageReader opens the chart package byte stream at u for digesting.
+// On success, it returns an [io.ReadCloser]: for http or https, the response body from [http.Get];
+// for file or an empty scheme, the file from [os.Open] when the URL path ends with ".tgz".
+// It returns (nil, nil) for file or an empty scheme when the path does not end with ".tgz".
+// It returns (nil, err) if the scheme is unsupported, or if [http.Get] or [os.Open] fails.
```
komish: This feels redundant, seeing it both in the exported Function as well as here, even though they have subtle differences. Can we figure out a way to make sure the comments don't overlap as heavily?

komish: Also, feels a little too specific by providing exact error case tuples as references. Better to capture the behavior at a higher level, and leave the return cases as something that can be observed by reading the logic itself, given it isn't very long.

internal/chartverifier/reportBuilder.go
```diff
@@ -223,30 +223,45 @@ func GenerateSha(rawFiles []*helmchart.File) string {
...
+func GetPackageDigest(uri string) string {
+	u, err := url.Parse(uri)
+	if err != nil {
+		return ""
 	}
-	if err != nil || chartReader == nil {
+	rc, err := openChartPackageReader(u)
+	if err != nil || rc == nil {
 		return ""
 	}
```
komish: I realize this is the original code, but unless you see a reason not to - let's flatten these out to be on separate conditions. E.g. remove the `||`

```
if err != nil { ... }
if rc == nil { ... }
```

komish (changes requested):
Few comments. Overall, I'm not sure if we're getting too much here given we don't really enrich or use any of the errors that would be returned by the new function, but if you believe it's more readable, we can merge this after the requested changes.

internal/chartverifier/reportBuilder.go
caxu-rh: I've shortened/condensed the godoc in both functions.

caxu-rh:
Thanks for the feedback. The goal was originally to ensure that reader (file or response body) gets closed, but due to the early-return style it would've required multiple `defer` in the lines above, hence the larger refactor. If it feels like too much, we can still keep the original, just need to have some more in each `case`.

komish (approved):
/lgtm

PR #620, opened 2026-06-11
Migrate from helm/v3 to helm/v4

komish:
This PR migrates to the Helm/v4 APIs (from Helm/v3). Where possible, the new library was swapped in while attempting to preserve as much of the original functionality as possible. I don't see any major breaking changes impacting Chart Verifier as a result of the migration.

Notes for maintainers:

- This migration does not include additional logic to support OCI-based chart URIs.

- As of this writing, Helm has not exported the structs for Chart version 3. This PR does not change our existing check for Chart version 2 ("used in Helm/v3") or make any changes to support Chart version 3. When this is published, we'll need to add additional logic and potential refactor further to support it.

-  Some packages shifted in import paths to support versioned APIs. E.g. `helm.sh/helm/v3/chart` --> `helm.sh/helm/v4/chart/v2` - this is a "breaking change", but generally only "breaking" because the import path isn't 1 to 1. If I recall correctly, accounting for the import path gave us the original functionality.

- Where relevant, I tried to use appropriate import path aliases to clarify what we're working with (e.g. `helm.sh/helm/v4/chart/v2` has a package name `v2` and should be aliased as `chartv2`, making room for a future `chartv3`.

- Helm previously had some functions that accepted logging functions, which we leveraged. That API has changed to support a slog.Handler, so I've added on here to maintain the original logging API used in Chart Verifier. I have a future patch partially prepared to migrate Chart Verifier to slog from the standard library.

- Helm/v4 provides some new functions for working with "Release" and "Chart" resources that we're introducing here. The design of these APIs is a little bit tricky, and hides some functionality behind interfaces that we otherwise used directly in structs. Typically, this is done to give users a stable API for resolving to the correct version of a given chart. You'll see these changes mostly in `RenderManifests`. I preferred to maintain the existing logic for now. We can revisit if we'd like to change the way `RenderManifests` works fundamentally at a later time.


internal/chartverifier/utils/slog_adapter.go
```diff
@@ -0,0 +1,90 @@
...
+// disabling the slog.Handler.
+func (h *slogHandler) Suppress(b bool) {
+	h.suppressed = b
+}
+
+func (h *slogHandler) IsSuppressed() bool {
+	return h.suppressed
+}
+
+func (h *slogHandler) Enabled(ctx context.Context, level slog.Level) bool {
```
mgoerens: We don't need either of the arguments.

pkg/chartverifier/report/types.go
```diff
@@ -3,7 +3,7 @@ package report
 import (
 	"net/url"
 
-	helmchart "helm.sh/helm/v3/pkg/chart"
+	helmchart "helm.sh/helm/v4/pkg/chart/v2"
```
mgoerens: Should the alias become `chartv2` to stay consistent with this import in other places ?

pkg/chartverifier/reportsummary/types.go
```diff
@@ -1,7 +1,7 @@
 package reportsummary
 
 import (
-	helmchart "helm.sh/helm/v3/pkg/chart"
+	helmchart "helm.sh/helm/v4/pkg/chart/v2"
```
mgoerens: Should the alias become `chartv2` to stay consistent with this import in other places ?

mgoerens (commented):
Thanks @komish for your work ! A few small comments.

Also, CI is failing and I don't think that is expected.

pkg/chartverifier/reportsummary/types.go
komish: This should be fixed now.

pkg/chartverifier/report/types.go
komish: This should be fixed now.

internal/chartverifier/utils/slog_adapter.go
komish: I've underscored these arguments, but we need the function signature to remain in order to match the slog.Handler interface (IIRC)

komish:
The recent pushes should include fixes to the `Uninstall()` workflow that was causing it to fail prematurely due to not having a configured wait strategy on the Uninstall client.

Also fixes a bug in the slogging adapter where we would not properly pass through mutexes and byte buffers with derivative loggers (e.g. those created via the With* methods) were used.

mgoerens (approved):
/lgtm

PR #632, opened 2026-07-09
Use legacy install and watch strategies from helm v3

komish:
Helm v4 introduces new default approaches for how charts are installed and checked for readiness, but these require additional permissions from the service account used to install.

We use a restricted service account in public clusters. Switching to the historical install and watch approach allows the service account in our pipeline to continue working with existing permissions. The newer approach requires more permissions than we want to give today (PATCH, LIST for resources at a cluster scope like ClusterRoleBindings).

This patch reverts the behavior to be consistent with history so that we can release with the helm v4 APIs in place, and allows us to consider the install and readiness check approaches at a later date, potentially with a larger version increment.

### DIFF
--- a/internal/tool/helm.go
+++ b/internal/tool/helm.go
@@ -43,10 +43,29 @@ func (h Helm) Install(ctx context.Context, namespace, chart, release, valuesFile
 	client := action.NewInstall(h.config)
 	client.Namespace = namespace
 	client.ReleaseName = release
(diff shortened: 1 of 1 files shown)


komish:
For reviewers: I expect this to fix issues with CI nightlies that are caused by chart verifier's inability to install certain charts in our test fixtures.

jsm84 (approved):
LGTM

PR #637, opened 2026-07-15
Group codeql action updates

komish:
Dependabot-generated PRs seem to be failing, and the reason may be because the various steps have different versions


jsm84 (approved):
LGTM

